PR #2, opened 2021-05-25
Making delete work

freshteapot:
# What
- Lock down edit / delete to tenants which can.

# Longer version
- We now check the git repo for a studio.json file which will have studioconfig, to give us a chance to expose the site but also lock it down.


freshteapot (commented):
Good enough

PR #7, opened 2021-06-08
Webhook to write to raw data log

freshteapot:
# Todo
- [x] Topic name
- [x] Define how to read the uri

# Reference
- https://app.asana.com/0/0/1200420198454049/f


freshteapot:
Good enough for now

PR #10, opened 2021-07-26
Exposing insights from the runtime

freshteapot:
# What
- endpoint to get some crude insights into the runtime for a microservice
- An endpoint to proxy requests to the log infrastructure (loki)

## Insights
- Latest event per event type
- Latest event
- Total events in log
- Current state of each event handler, including retry informaiton
- Current state of each event filter

# Information
- requires more network policy access to access mongo within each namespace


freshteapot (commented):
First cut doesn't look to bad. It might be too "internal" the api.

But its a start.

PR #15, opened 2021-09-03
Functionality to create and delete Purchase order api Microservice kind

woksin:
## Summary

Adds the functionality to create and delete purchase order api microservices

### Added

- Input types for PurchaseOrderAPI microservice for when doing a request to create a microservice
- MicroserviceKind for PurchaseOrderAPI
- Create purchase order api microservice
- Delete purchase order api microservice

### Fixed

- Some issues in regard to raw data log
- Bug related to MongoDB connection strings


pkg/platform/doc.go
```diff
@@ -105,22 +119,30 @@ type GitRepo interface {
 	GetAll(tenantID string) ([]Application, error)
 }
 
-const (
-	Simple                 = "simple"
-	BusinessMomentsAdaptor = "business-moments-adaptor"
-	RawDataLogIngestor     = "raw-data-log-ingestor"
-)
+var MicroserviceKinds struct {
```
freshteapot, line 122: The code you replaced, is simpler to read.
Please revert.

freshteapot (changes requested):
I am struggling a little to understand why you have refactored the code.
The work needed to do, was reasonably self contained.

My concern with your approach.

- You haven't actually created the feature we want in this PR as of now.
- You haven't written tests before you refactored.
- We can debate if your refactor is of benefit in terms of readability and or maintainability.

I have watched you refactor https://github.com/dolittle/ERPIntegrations/ along with others, since refactoring, even tho you were happy to change the code, when asked about why you hadn't changed to embeddings, you blamed the original writer of the code. Which has me concerned your not fully embracing owning the code, even tho your willing to heavily modify / refactor / move code around.

I am all for improving the code base, but as mentioned in the retrospective "product should trump tech". This conversation would be very different if the functionality was implemented and tests existed (not your fault they didn't exist before you jumped in the code).

pkg/platform/doc.go
jakhog: A nice way to do enumeration-like things in Go could be:
```go
type MicroserviceKind string

const (
  MicroserviceKindSimple MicroserviceKind = "simple"
  MicroserviceKindBusinessMomentsAdaptor  = "business-moments-adaptor"
  MicroserviceKindRawDataLogIngestor      = "raw-data-log-ingestor"
  MicroserviceKindPurchaseOrderAPI        = "purchase-order-api"
)
```

woksin: Hmm, better than that ugly struct mess. Though I liked how it was used (platform.Kind.*), relying on the naming is also fine

pkg/platform/microservice/businessmomentsadaptor_repo.go
```diff
@@ -133,214 +129,86 @@ func (r businessMomentsAdaptorRepo) Create(namespace string, tenant k8s.Tenant,
...
 	return nil
 }
 
 func (r businessMomentsAdaptorRepo) Delete(namespace string, microserviceID string) error {
-	client := r.k8sClient
 	ctx := context.TODO()
-	// Not possible to filter based on annotations
-	opts := metaV1.ListOptions{}
-	deployments, err := client.AppsV1().Deployments(namespace).List(ctx, opts)
+	deployment, err := k8sGetDeployment(r.k8sClient, ctx, namespace, microserviceID)
```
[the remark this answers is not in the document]
woksin: Your're right. I'll do that

pkg/platform/doc.go
goeran: @freshteapot It would be great if you could have offered an explanation of why you think the code that was replaced was easier to read. That would have given the author and other reviewers (me) a chance to learn from you.

freshteapot: Sure @goeran :).


# A
- Not against the prefix above (just using the code from master)
```go
const (
	Simple                 = "simple"
	BusinessMomentsAdaptor = "business-moments-adaptor"
	RawDataLogIngestor     = "raw-data-log-ingestor"
)
```

VS

# B
```go
var MicroserviceKinds struct {
	Simple                 string
	BusinessMomentsAdaptor string
	RawDataLogIngestor     string
	PurchaseOrderAPI       string
} = struct {
	Simple                 string
	BusinessMomentsAdaptor string
	RawDataLogIngestor     string
	PurchaseOrderAPI       string
}{
	Simple:                 "simple",
	BusinessMomentsAdaptor: "business-moments-adaptor",
	RawDataLogIngestor:     "raw-data-log-ingestor",
	PurchaseOrderAPI:       "purchase-order-api", // TODO purchase-order-api VS purchase-order
}
```
- Scanning the code, your eyes will take in A much quicker, than B, because there is less syntax to "process".
- using platform.Kind.* etc is not a style I have seen in golang. Which I am speculating will make learning the code for a go developer or non go developer different from what they are used to or learning.
- Does seeing simple 3 times in 5 lines, seem right to you?

If I flip the question around:

- We needed to add a new kind, but why did we not follow the pattern that was already there?
- Did changing the pattern that was already there, improve the maintainability of the code?

jakhog: For reference, this was my inspiration with the prefix:
https://pkg.go.dev/net/http#pkg-constants

The custom type thing is just a nice to have.

woksin: Yes the struct pattern was pretty shitty to look at and extend. I liked @jakhog's pattern because it satisfies the strongly typed enum type which was what I was looking for

woksin:
Yes the functionalities are not implemented, that's why this is a draft PR

woksin:
This should be a minor, @joelhoisko  :)

joelhoisko:
@woksin whoops sorry, I was editing the wrong PR

freshteapot (commented):
Removing my "Request Changes".

PR #18, opened 2021-09-07
Fix dealing with Windows filepaths

joelhoisko:
## Summary

Fixes the filepath code to also work with Windows.

### Fixed

- Fixes the code to also understand Windows style path delimiters instead of just *nix paths when reading the repo.

### DIFF
--- a/pkg/platform/storage/git/application.go
+++ b/pkg/platform/storage/git/application.go
@@ -126,7 +126,7 @@ func (s *GitStorage) GetApplications(tenantID string) ([]platform.HttpResponseAp
 	for _, applicationID := range applicationIDs {
 		application, err := s.GetApplication(tenantID, applicationID)
 		if err != nil {
-			fmt.Println(err)
+			fmt.Println("Skipping application for tenant", tenantID, "with ID", applicationID, "because it failed to load:", err)
 			continue
 		}
 		applications = append(applications, application)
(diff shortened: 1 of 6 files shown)


pkg/platform/storage/git/application.go
```diff
@@ -136,6 +126,7 @@ func (s *GitStorage) GetApplications(tenantID string) ([]platform.HttpResponseAp
 	for _, applicationID := range applicationIDs {
 		application, err := s.GetApplication(tenantID, applicationID)
 		if err != nil {
+			fmt.Println(err)
```
jakhog, line 129: Maybe nice with a little context for this error:
```suggestion
			fmt.Println("Skipping application for tenant", tenantID, "with ID", applicationID, "because it failed to load:", err)
```

jakhog (commented):
Nice, the changes look good to me. A little suggestion on a printing of an error.
Not sure as you say if there are more things that needs same treatment, but we can fix those later.

joelhoisko:
Just for archives sake, here's my original PR text:

The filepaths only assumed *nix style separators `/`, so the code failed to find the applicationID and to read the `application.json` file. This PR fixes it by using the proper `path/filepath` package, which deals with paths in a portable way.

I know that there are still some wonky spots, like we do something like this, but I'm not really sure whats going on here so I'll just ignore them for now (couldn't recreate these things in my tests either, as they seem to all be dealing with the creation of new applications/environments, which we don't have the UI for yet).
```go
err = w.AddWithOptions(&git.AddOptions{
	Path: strings.TrimPrefix(filename, s.Directory+"/"),
})
```

PR #19, opened 2021-09-09
Change git storage applications structure 

woksin:
## Summary

Changes the structure of the HttpInputEnvironment struct so that it knows about the tenants and has a mapping between tenant Id and ingress.

There are some assumptions in play around the ingresses that are added to the map. First it assumes that the ingresses that needs to be configured has one or more spec.TLS items. Secondly it will override the ingress config for a tenant. It also maybe clunkily extracts the tenant id from the ingress from the nginx.ingress.kubernetes.io/configuration-snippet annotation. 

### Changed

- Changed structure of HttpInputEnvironment to include a list of tenants and a mapping between tenant id and ingress host and domain prefix
- build-applications command to store applications with this new structure

# Reference
- https://app.asana.com/0/1200879032451150/1200931738686043/f

### DIFF
--- a/cmd/microservice/build-applications.go
+++ b/cmd/microservice/build-applications.go
@@ -88,7 +88,9 @@ func extractApplications(ctx context.Context, client *kubernetes.Clientset) []pl
 	applications := make([]platform.HttpResponseApplication, 0)
 
 	for _, ns := range getNamespaces(ctx, client) {
-		// Creates a single application for a namespace(?)
+		if !strings.HasPrefix(ns.GetObjectMeta().GetName(), "application-") {
+			continue
+		}
 		applications = append(applications, getApplicationFromK8s(ctx, client, ns))
 	}
 
--- a/cmd/microservice/build-applications.go
+++ b/cmd/microservice/build-applications.go
@@ -242,7 +242,7 @@ func environmentAlreadyInApplication(environments []platform.HttpInputEnvironmen
 		return item.Name == environmentName
 	})
 
-	if index != -1 {
+	if index == -1 {
 		return false, index
 	}
 	return true, index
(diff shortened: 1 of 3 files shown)


cmd/microservice/build-applications.go
```diff
@@ -79,73 +86,171 @@ func SaveApplications(repo storage.Repo, applications []platform.HttpResponseApp
...
+	}
+	return ingresses.Items
+}
+
+func getTenantsFromConfigmap(ctx context.Context, client *kubernetes.Clientset, namespace, environmentName string) map[string]interface{} {
+	tenantsConfig, err := client.CoreV1().ConfigMaps(namespace).Get(ctx, fmt.Sprintf("%s-tenants", strings.ToLower(environmentName)), metaV1.GetOptions{})
+	if err != nil {
+		panic(err.Error())
+	}
+	tenantsJson := tenantsConfig.Data["tenants.json"]
```
woksin, line 218: I'm a bit unsure about this part, how to extract the data. Is this the correct way?

jakhog: Yep

woksin:
Doesn't look like this will get in today, then I'll see that as an oppurtunity to make some specs

joelhoisko:
We're fixing some bugs on it now too

jakhog:
I've added a field in the EnvironmentIngress to also store the secret name (which we will need when creating a RawDataLog or updating the webhooks). Also simplified the build-applications code a little bit since we can now rely on the "tenants.json" ConfigMap to retrieve environments.

jakhog:
Otherwise it LGTM. I've tested this locally with the production cluster, and it runs just fine generating all the JSON files.

jakhog:
I've also made a PR for the corresponding update of the local sample data here https://github.com/dolittle/Studio/pull/106

PR #20, opened 2021-09-10
Verify created pod labels and add prefix to name for RawDataLog

jakhog:
## Summary

I've changed the setup of NATS+STAN resources in the RawDataLogRepo:
1. The name now includes the environment (to allow more than one)
2. Removed the pod-anti-affinity settings on the STAN statefulset so that we can have more than 3 in total
3. Added the Dolittle labels to the statefulsets (and pod templates) so that they are covered by the network policies
4. Changed the setup so they are created with the non-dynamic Kubernetes client (to make it easier to test - the dynamic client requires a connection to a cluster to work).

The tests are quite big, but I'm not verse with Ginkgo yet, so improvements on the structure are welcome :)

I tried this out locally, and it does not work with the k3d cluster (the nats pod doesn't allow any connections). But with the docker cluster it works.

#### Asana tasks
- https://app.asana.com/0/1200747172416688/1200762338877903/f
- https://app.asana.com/0/1200879032451150/1200931738686016/f

### Changed

- The creation of NATS+STAN resources now includes environment in the name, has labels to work with existing networkpolicies
- Change creation of RawDataLog for NATS+STAN resources to use the non-dynamic (hardcoded) Kubernetes client.
- Changed `*kubernetes.ClientSet` to `kubernetes.Interface` everywhere to enable the use of the Fake client.

### Fixed

- The wrong tenant id was given when creating config map for rawdatalogingestor microservice


jakhog:
Turns out it is not k3d's fault it doesn't work. So I'll need to do a bit more digging.

jakhog:
Now it works locally with k3d - turns out it was a mixture of wrong labels that made the NetworkPolicy for the microservice apply to everything (essentially disallowing all traffic). Thanks @freshteapot

pkg/platform/microservice/rawdatalog/repo.go
```diff
@@ -64,53 +47,33 @@ func (r RawDataLogIngestorRepo) Update(namespace string, tenant k8s.Tenant, appl
...
-		"environment":  input.Environment,
-		"microservice": input.Name,
+		"tenant":      tenant.Name,
+		"application": application.Name,
+		"environment": input.Environment,
 	}
 
 	annotations := map[string]string{
-		"dolittle.io/tenant-id":       customer.ID,
+		"dolittle.io/tenant-id":       tenant.ID,
```
freshteapot, line 59: We have GetLabels and GetAnnotations in https://github.com/dolittle/platform-api/blob/58acca47238064bcc5629db6dfa373f2eea8277f/pkg/dolittle/k8s/shared.go

If not, we just need to decide if we want to include 
> "dolittle.io/microservice-kind": string(microservice.Kind),

which got introduced, to help give context to the microservice, when talking only in kubernetes.

pkg/platform/microservice/rawdatalog/nats.go
```diff
@@ -0,0 +1,231 @@
...
+				},
+				{
+					Name:       "metrics",
+					Port:       7777,
+					TargetPort: intstr.FromString("metrics"),
+				},
+				{
+					Name:       "leafnodes",
+					Port:       7422,
+					TargetPort: intstr.FromString("leafnodes"),
```
freshteapot, line 77: Should this point to 4222 like it does in the "pure yaml" version?

freshteapot (approved):
Aside from my comments, which im open to input or being ignored.
It would be nice to remove the existing k8s files.

ie
- pkg/platform/microservice/rawdatalog/k8s/*
- k8s/*

Otherwise, I can do that afterwards in another PR.

pkg/platform/microservice/rawdatalog/nats.go
jakhog: Done

pkg/platform/microservice/rawdatalog/repo.go
jakhog: I've changed the code to use them, but I'm not sure it's really correct. Especially the Microservice Name+ID (I think they should both be fixed)

jakhog:
I've deleted the k8s files, and done the suggested changes. But the code doesn't build just now (because the GitStorage was added to the constructor, which I haven't done in the tests).

I don't have time today or tomorrow to fix it, but would be nice with a fake/mock implementation of the storage that doesn't do anything that we could use for this test.

I also think that the test will fail once they build, because we need to not have a "microservice" label on the NATS/STAN deployments for the NetworkPolicies to not kick in. That should be a simple fix though (just remove them from the map after calling the standard functions).

Lastly (also in comment above), I think the Microservice Name+ID should be fixed for the rawdatalogingestor. Should we do that in the backend (and remove from input), or just put it in the frontend?

woksin:
I went back on a commit I did and did this https://github.com/dolittle/platform-api/pull/20/commits/aee1133bff4d73a0e798936cfe19b92c13ddaae0

I realised that I myself got confused over what the tenants really are. As we know this would be a lot simpler when we get to change the names. Customer and CustomersTenant?

woksin (approved):
Lots of stuff, but to my knowledge this looks good to me

PR #22, opened 2021-09-14
Use configured tenant when creating POAPI and RawDataLogIngestor

woksin:
## Summary

When creating Purchase Order API microservice or RawDataLogIngestor microservice it is important to configure it to the correct tenant. The tenant being used before was just a static tenant, it needs to not be static in order for it to work with actual customers.

There are a few caveats at the moment:
- It will only use one tenant. For instance if we needed a Purchase Order API to commit events to multiple tenants that wouldn't work with the current solution
- The added configured tenant is not used for the other microservices yet. Only used for Purchase Order API and RawDataLogIngestor
- When we want to do this for normal microservices we need to change this to configure for all tenants, not just the first one 

### Added

- A method for getting tenant for environment on the application metadata stored in git
- Specs for method for getting tenant for environment on the application metadata stored in git 

### Changed

- The tenant used when creating Purchase Order API and RawDataLogIngestor
- A couple of places in the code 'tenant' parameter was renamed to 'customer' to make it more explicit that the tenant actually represents a Dolittle customer

### DIFF
--- a/pkg/platform/microservice/purchaseorderapi/k8sResourceSpecFactory.go
+++ b/pkg/platform/microservice/purchaseorderapi/k8sResourceSpecFactory.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/dolittle-entropy/platform-api/pkg/dolittle/k8s"
 	"github.com/dolittle-entropy/platform-api/pkg/platform"
-	microserviceK8s "github.com/dolittle-entropy/platform-api/pkg/platform/microservice/k8s"
 	corev1 "k8s.io/api/core/v1"
 )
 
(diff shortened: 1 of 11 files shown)


woksin:
I was a little unsure on how I wanted to implement the functionality for getting the configured tenant. I know I wanted to write specs for it so the options I had in mind was to:
- Add methods to the stored application struct (as I did)
- Add functions that take in the stored application struct as a parameter

I'm open to suggestions if there are better ways of doing this

pkg/platform/microservice/purchaseorderapi/k8sResourceSpecFactory.go
```diff
@@ -17,24 +16,24 @@ func NewK8sResourceSpecFactory() K8sResourceSpecFactory {
...
 
-func (r *k8sResourceSpecFactory) modifyEnvironmentVariablesConfigMap(environmentVariablesConfigMap *corev1.ConfigMap, k8sMicroservice k8s.Microservice, extra platform.HttpInputPurchaseOrderExtra) {
-	resources := k8s.NewMicroserviceResources(k8sMicroservice, microserviceK8s.TodoCustomersTenantID)
-	mongoDBURL := resources[microserviceK8s.TodoCustomersTenantID].Readmodels.Host
-	readmodelDBName := resources[microserviceK8s.TodoCustomersTenantID].Readmodels.Database
+func (r *k8sResourceSpecFactory) modifyEnvironmentVariablesConfigMap(environmentVariablesConfigMap *corev1.ConfigMap, k8sMicroservice k8s.Microservice, tenant platform.TenantId, extra platform.HttpInputPurchaseOrderExtra) {
+	tenantID := string(tenant)
+	resources := k8s.NewMicroserviceResources(k8sMicroservice, tenantID)
+	mongoDBURL := resources[tenantID].Readmodels.Host
+	readmodelDBName := resources[tenantID].Readmodels.Database
```
joelhoisko, line 35: Not all `resources.json` files have the `readModels` key, it's a semi-legacy thing from the past and it mostly exists because usually the head code also reads the same `resources.json`, but that is just a convention and not enforced by anything really. But this is also touching upon the fact that not all customers applications have a read model store. The closest "official" thing would be our projection/embedding store resource definitions. We could also just get the `mongoDBURL` from the event store definition and then create the `readmodelDBName` from something.

For now this'll do, but we have to think of something in the future.

pkg/platform/microservice/service.go
```diff
@@ -78,23 +78,23 @@ func (s *service) Create(responseWriter http.ResponseWriter, request *http.Reque
...
 		return
 	}
 
-	if !s.gitRepo.IsAutomationEnabled(tenant.ID, applicationID, environment) {
+	if !s.gitRepo.IsAutomationEnabled(customer.ID, applicationID, environment) {
 		utils.RespondWithError(
 			responseWriter,
 			http.StatusBadRequest,
 			fmt.Sprintf(
 				"Tenant %s with application %s in environment %s does not allow changes via Studio",
```
joelhoisko, line 96:
```suggestion
				"Customer %s with application %s in environment %s does not allow changes via Studio",
```

Should we also change it to "Customer" in here? We've really muddied the waters with different types of tenants 😂

joelhoisko (changes requested):
Bueno stuff, one real suggestion/point and one point about how the `resources.json` in the platform are a bit messy.

pkg/platform/microservice/purchaseorderapi/k8sResourceSpecFactory.go
woksin: Yeah, it has been like that so I did not want to touch it.

pkg/platform/microservice/service.go
jakhog: I vote for gradually changing the "tenants of dolittle" -> "customers". As Joel said, it's a bit tricky to always have to specify which kind of tenant we're talking about in conversations and in the code. I think it would be too big of a job to fix all references right away, but changing as we come across it is fine by me.

woksin: I agree.  Then I'll commit the suggestion

joelhoisko (approved):

PR #24, opened 2021-09-15
Error if poapi exists

woksin:
## Summary

Checks whether a purchase order api microservice already exists by comparing the name of the deployment If it already exists it will return with an error and write a response with StatusInternalServerError

### Added

- Check for whether purchase order api exists when creating it

### Changed

- Changed some parameters and variables from *kubernetes.ClientSet to kubernetes.Interface in order to write specs
- Temporarily removed the dependency to rawDataLogIngestorRepo from purchaseorderapi repo in order to make it easier to write specs


pkg/platform/microservice/purchaseorderapi/repo.go
```diff
@@ -53,6 +57,33 @@ func (r *repo) Create(namespace string, customer k8s.Tenant, application k8s.App
...
+		Tenant:      customer,
+		Application: application,
+		Environment: environment,
+		ResourceID:  microserviceK8s.TodoCustomersTenantID,
+		Kind:        platform.MicroserviceKindPurchaseOrderAPI,
+	}
+
+	ctx := context.TODO()
+
+	resources := r.k8sResourceSpecFactory.CreateAll(headImage, runtimeImage, microservice, tenant, input.Extra)
```
woksin, line 80: This is a heavy-handed way of getting the deployment name. Maybe we should consider at some point to make it more streamlined? Since there is only one way to create deployment names for microservices

pkg/platform/microservice/purchaseorderapi/repo.go
```diff
@@ -49,6 +53,33 @@ func (r *repo) Create(namespace string, customer k8s.Tenant, application k8s.App
 
 	return nil
 }
+func (r *repo) Exists(namespace string, customer k8s.Tenant, application k8s.Application, tenant platform.TenantId, input platform.HttpInputPurchaseOrderInfo) (bool, error) {
+	// TODO not sure where this comes from really, assume dynamic
```
joelhoisko, line 57:
```suggestion
```

I guess this should be removed, we do now know where it comes from

joelhoisko (changes requested):
👍

PR #25, opened 2021-09-16
Filter applications

woksin:
## Summary

When getting applications from git repo start walking the directory from the customer folder, instead of the root of the repo.

### Changed

- When getting applications from git repo start searching from the customer folder
- Use the logger


jakhog:
@woksin I changed the log printing to use the structured log fields and converted a few other fmt.Print -> log.- statements. Wasn't sure I would get the syntax right, so didn't dare do it with a suggestion - sorry. If you're unhappy, just remove the two latest commits.

woksin:
I approve of these changes

joelhoisko (approved):
👍

PR #26, opened 2021-09-16
Create RawDataLog, Stan & Nats when creating POAPI if they don't already exist

joelhoisko:
## Summary

When creating a PurchaseOrderAPI microservice, if the environment doesn't already have a RawDataLog microservice and Stan & Nats it will create them.

### Added

- Creates the missing RawDataLog deployment and Nats & Stan statefulsets when creating a PurchaseOrderAPI microservice if they didn't already exist
- Save the RawDataLog microservice to the git repo
- Add logging for the creation of RawDataLog

### Changed

### Fixed

- Fixes the `NATS_CLUSTER_URL` env to the hardcoded `<env>-nats.namespace...` format
- Also makes the RawDataLog handler (the specific one) to check for the RawDataLogs existence beforehand


joelhoisko:
Built on top of https://github.com/dolittle/platform-api/pull/20. Tested to work with Studio https://github.com/dolittle/Studio/pull/104 branch

woksin (approved):
This seems correct to me. Other than I would personally prefer it to be a bit cleaner it is good enough 👍

PR #27, opened 2021-09-16
Little cleanup

woksin:
## Summary

Went on a little personal journey to learn more about golang. I think that the result is a more streamlined way to handle platform/microservice requests (Create/Delete). Some bugs that went under the radar previously has also been fixed

### Changed

- Changed the handling of microservice requests to be more stream lined and less prone to errors

### Fixed

- Some errors that were not handled
- Created purchase order api twice


woksin:
I did most of this late last night as a journey into becoming better at golang. It's not super important to land this (though I prefer the cleanliness it brings :)) but I did however fix a couple of bugs

PR #29, opened 2021-09-20
Change the configured http port for POapi

jakhog:
## Summary

Fixes the configured HTTP listen on port when deploying a PurchaseOrderAPI microservice, so that it conforms to the convention in the platform (to listen on port 80 for the ingress). Even though we don't have an ingress automatically set up for it now, this makes the pod listen on the port that the service expects it to. So that when we do want to create an ingress (to port 'http') this will work.

I have tested this locally with k3d, and it works.

### Fixed

- Change the created environmental variables for the PurchaseOrderAPI deployment so that it listens on port 80 instead of 8080


freshteapot (approved):
Looks good.
I like that this is explicit.

I had looked a little at the code earlier to see if we could hook it upto the golang k8s object, but that would require more work.

PR #30, opened 2021-09-20
Only allow single PurchaseOrderAPI per environment

jakhog:
## Summary

Adds checks to ensure that there doesn't already exist a PurchaseOrderAPI in an environment when creating. The previous logic checked for one with the same name and id, the added code checks for any name and id.

The code might be a bit excessive, as it now first checks the Git storage, and then K8s afterwards. We could get away with one of them, but I'm not sure which one to truly consider the source of truth until we have everything covered by automation. Also might not be clear, but the old check is still useful because it checks for any microservice with the same name (which would cause problems) - even though the naming of the function kind of indicates that it looks for a PurchaseOrderAPI microservice.

I've tested it locally, and also added more specs for the `Exists` method and the new `EnvironmentHasPurchaseOrderAPI` method.

### Added

- Disallow the creation of multiple PurchaseOrderAPIs in the backend.

### Changed

- The structure of the specs for the purchase order api repo

### DIFF
--- a/.github/workflows/check_labels.yml
+++ b/.github/workflows/check_labels.yml
@@ -16,5 +16,5 @@ jobs:
     steps:
       - uses: docker://agilepathway/pull-request-label-checker:latest
         with:
-          one_of: major,minor,patch
+          one_of: major,minor,patch,no-release
           repo_token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/check_labels.yml
+++ b/.github/workflows/check_labels.yml
@@ -1,0 +1,20 @@
+name: Verify release labels
+
+on:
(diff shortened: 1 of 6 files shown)


jakhog:
Oh, let me have a quick look at the tests if there is anything I should add there.

joelhoisko (approved):
Nice stuff, my only point would be that it would be nice to have logging whenever we return an error to the client. But I also know that we've been thinking about doing some sort of collective thinking around how we want to do logs in general so I'm okay with letting it wait for now.

jakhog:
I'll add some logs in the Handler

joelhoisko (approved):
👍

PR #33, opened 2021-09-24
Return "N/A" for start-time when pod is starting up

jakhog:
## Summary

While a pod I starting up, the StartTime is not set (nil in Go) which caused the "View" screen just after creating e.g. a POapi to fail. This fix just returns "N/A" instead of an actual date when StartTime is nil. The contract was just a plain string anyways, so Studio will accept it.

#### How to test
- Start with a clean cluster (or without raw-data-log and purchase-order)
- Click Microservices
- Click Create new microservice
- Click Create within the Purchase order api card
- Select Infor graphic
- Enter name
- Click next
- Enter username
- Enter password
- Click Next
- At this point, we send the request to the backend and redirect the frontend.
- Observe platform-api __does not__ crash

### Fixed

- Handling of Pod.StartTime nil-values while the pod is still starting up.


woksin (approved):
I have tried it out locally myself and it seems to work

PR #35, opened 2021-09-27
Raw data log microservice bug

woksin:
## Summary

There was a bug in the raw data log ingestor repo implementation that converted the environment from the microservice from the input to lower case, causing the environment label on the raw data log deployment to be different from the other microservice kind deployments. 

However this fix has a side effect that I don't know the effects of, since we do not convert the environment to lower that results in the DOLITTLE_ENVIRONMENT environment variable on the envVariables config map being different (not lowercased). Is this important? If so we can just make the input there to lower case, that's not a big deal.
EDIT: As of writing this the code in question is on line 444

#### How to test
- Go through test steps in PR in Studio https://github.com/dolittle/Studio/pull/116

### Fixed

- Deployment of raw data log should now have the environment label in the proper casing

### DIFF
--- a/pkg/platform/microservice/rawdatalog/repo.go
+++ b/pkg/platform/microservice/rawdatalog/repo.go
@@ -372,7 +372,7 @@ func (r RawDataLogIngestorRepo) doDolittle(namespace string, customer k8s.Tenant
 		"method":    "RawDataLogIngestorRepo.doDolittle",
 	}).Debug("Starting to create RawDataLog microservice")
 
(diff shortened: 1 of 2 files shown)


joelhoisko (approved):
Looks good, I also checked the code for RawDataLog command and it doesn't use the `DOLITTLE_ENVIRONMENT` variable right now for anything (and the default value given to it is also in lower case).

joelhoisko:
Tested, merging this in 👍

PR #36, opened 2021-09-28
Make STAN deployment persistent

joelhoisko:
## Summary

Adds a PersistentVolumeClaim to the STAN deployment with the name of `<env>-stan-storage` for the PVC. It uses the same [`managed-premium`](https://docs.microsoft.com/en-us/azure/aks/concepts-storage#storage-classes) storage class as we use for MongoDB in the platform. The STAN store method is [file](https://docs.nats.io/nats-streaming-server/configuring/persistence/file_store) and the name of the directory is `datastore` like in the STAN doc examples.

To use this locally you have to create a new storage class with the name `managed-premium`, create a persistent volume attached to it, and attach a directory to that persistent volume during k3d creation. This is documented in the Studio PR https://github.com/dolittle/Studio/pull/119

### Added

- Adds a PersistentVolumeClaim of 8GB to the STAN deployment

### Changed

- Made some logging around NATS/STAN creation more clear in RawDataLog

### DIFF
--- a/pkg/platform/microservice/rawdatalog/stan.go
+++ b/pkg/platform/microservice/rawdatalog/stan.go
@@ -2,9 +2,11 @@ package rawdatalog
 
 import (
 	"fmt"
+	"log"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/util/intstr"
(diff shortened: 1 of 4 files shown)


pkg/platform/microservice/rawdatalog/stan.go
```diff
@@ -19,6 +21,13 @@ type stanResources struct {
 func createStanResources(namespace, environment string, labels, annotations labels.Set) stanResources {
 	name := fmt.Sprintf("%s-stan", environment)
 	natsName := fmt.Sprintf("%s-nats", environment)
+	quantity, err := resource.ParseQuantity("8Gi")
```
joelhoisko, line 24: I don't know how big the PVC resource request should actually be, now I just made it the same `8Gi` that we do for Mongo by default. In Studio's yaml files I set the persistent volume to have a capacity of `10Gi` so that should cover the `8Gi` at least.

freshteapot (approved):
Nice work @joelhoisko.
I like how you created the persitentvolume for the local cluster.

I have confirmed this working locally.
I have also confirmed this working in the cluster.


I note in the production cluster we are setting up with
> persistentVolumeReclaimPolicy: Delete

https://kubernetes.io/docs/tasks/administer-cluster/change-pv-reclaim-policy/#why-change-reclaim-policy-of-a-persistentvolume

I don't think it matters for now, as we are living in a world where we don't want to let people delete the raw-data-log.
I also feel once we have backups in place, this is less an issue.

Just learning. I found it hard work to learn about this in azure.

joelhoisko:
@freshteapot Also just to learn more:

> I note in the production cluster we are setting up with
>> persistentVolumeReclaimPolicy: Delete

I tried to find any mention of this `persistentVolumeReclaimPolicy` in the production cluster, but I couldn't find anything about it. Or do you mean that by default they all have it set to Delete like the link mentions about "dynamically provisioned PersistentVolumes" (is that what we are doing)? Or is this some deeper default k8s configuration from AKS or something?

PR #37, opened 2021-09-29
Update raw data log webhooks

woksin:
## Summary

This PR adds the functionality to change the webhooks configuration for purchase order api and raw data log, both when creating/deleting purchase order api and updating the username/password on an already existing purchase order api (and by extension raw data log).

#### How to test (prerequisites)
- Start with a clean cluster (or without raw-data-log and purchase-order)
- Click Microservices
- Click Create new microservice
- Click Create within the Purchase order api card
- Select Infor graphic
- Enter name
- Click next
- Enter username
- Enter password
- Click Next
- Click Microservices
- See that both raw data log and purchase order api exists

##### Test that deleting and re-creating updates rawdatalog webhook
- Delete the purchase order api
- Recreate it following the steps above, but make sure to enter another username/password combination
- Verify that both raw data log and purchase order api exists
- Navigate to your local dev git storage (/tmp/dolittle-local-dev)
- Verify that the microservices for purchase order api and raw data log has the same "webhooks" object

##### Test that updating purchase order api updates rawdatalog webhook
- Use your favorite tool do a PUT request to /microservice with the correct headers and a body that is a purchase order api microservice input structure (the same as when you create one) with another value for the "authorisation" string on the webhooks objects
- Verify that both raw data log and purchase order api exists
- Navigate to your local dev git storage (/tmp/dolittle-local-dev)
- Verify that the microservices for purchase order api and raw data log has the same "webhooks" object with the "authorisation" string you had in the body

### Added

- PUT /microservice route that uses the Update request handler on the microservice.Service struct
- Update method in microservice.Service that currently only can update the webhooks of a purchase order api. In the future we should consider how to extend this method to other microservice kinds, this implementation was just done to achieve to shortest path to the immediate goal.
- UpdateWebhooks method on purchaseorderapi.Handler that updates only the webhook configuration of the existing purchase order api and raw data log. It will return error if purchase order api does not exist, but it will create the raw data log if that doesn't exist (which shouldn't ever be the case)
- Update method on rawdatalog.Repo that only updates the config files config map of the raw data log ingestor repo. I'm not certain that this will ensure that the configuration will be refreshed for the deployment, though after talking with @jakhog we think that it will work this way.

### Changed

- Changed server endpoints to use the http.Method* constants instead of magic strings
- Added (and use) the logger to the microservice.Service struct
- Purchase Order API handler signature so that it does not take in the response writer, but instead returns a custom Error type that includes the error and the http status code. This is done because it should be the responsibility of the caller (microservice.Service) to respond with a http response.
- Change the behaviour of the purchaseorderapi.Handler Create method so that it ensures that a purchase order api does not exist and also ensures that a raw data log ingestor exists with the configured webhoooks from the purchase order api microservice information from the request. Meaning that it will create a new raw data log if it does not exist or update the existing one with the webhook configuration form the request.
- Exists method on rawdatalog.Repo now returns the microservice ID as well
- 

### Fixed

### Removed

- requests package with RequestHandler interface since it is not needed

### Security

### Deprecated


joelhoisko:
Just moving your demoing steps out from the summary (don't think we want them in the release notes):

#### How to test (prerequisites)
- Start with a clean cluster (or without raw-data-log and purchase-order)
- Click Microservices
- Click Create new microservice
- Click Create within the Purchase order api card
- Select Infor graphic
- Enter name
- Click next
- Enter username
- Enter password
- Click Next
- Click Microservices
- See that both raw data log and purchase order api exists

##### Test that deleting and re-creating updates rawdatalog webhook
- Delete the purchase order api
- Recreate it following the steps above, but make sure to enter another username/password combination
- Verify that both raw data log and purchase order api exists
- Navigate to your local dev git storage (/tmp/dolittle-local-dev)
- Verify that the microservices for purchase order api and raw data log has the same "webhooks" object

##### Test that updating purchase order api updates rawdatalog webhook
- Use your favorite tool do a PUT request to /microservice with the correct headers and a body that is a purchase order api microservice input structure (the same as when you create one) with another value for the "authorisation" string on the webhooks objects
- Verify that both raw data log and purchase order api exists
- Navigate to your local dev git storage (/tmp/dolittle-local-dev)
- Verify that the microservices for purchase order api and raw data log has the same "webhooks" object with the "authorisation" string you had in the body

woksin:
Good decision, @joelhoisko

pkg/platform/microservice/purchaseorderapi/handler.go
```diff
@@ -55,140 +74,221 @@ func (s *RequestHandler) Create(responseWriter http.ResponseWriter, r *http.Requ
...
-		}
+		return s.createRawDataLog(msK8sInfo, ms, logger)
+	} else {
+		return s.updateRawDataLogWebhooks(msK8sInfo, ms.Extra.Webhooks, ms.Environment, microserviceID, logger)
 	}
-	// TODO: Since we only do ingress validation on the creation of RawDataLog, this means that if it exists - we don't do any validation.
+}
 
-	err = s.repo.Create(msK8sInfo.Namespace, msK8sInfo.Tenant, msK8sInfo.Application, tenant, ms)
+func (s *Handler) ensurePurchaseOrderAPIExists(msK8sInfo k8s.MicroserviceK8sInfo, ms platform.HttpInputPurchaseOrderInfo, tenant platform.TenantId, logger *logrus.Entry) *Error {
```
joelhoisko, line 206: I see that you have both an `ensurePurchaseOrderAPIExists()` method and a `ensurePurchaseOrderAPIDoesNotExist()` methods, could these be made into a single one? It could maybe return an extra boolean to show whether it exists or not? They also check for the existence in a different way, which feels a bit off.

pkg/platform/microservice/rawdatalog/repo.go
```diff
@@ -646,3 +690,22 @@ func (r RawDataLogIngestorRepo) getEnvironmentFromApplication(application platfo
...
+	container.ImagePullPolicy = "Always"
+	container.Args = []string{
+		"raw-data-log",
+		"server",
+	}
+	deployment.Spec.Template.Spec.Containers = []corev1.Container{
+		container,
+	}
+	deployment.Spec.Template.Spec.Containers[0].Ports[0].ContainerPort = 8080
+}
```
joelhoisko, line 711: Maybe it's just me and my simple knowledge of go, but I don't really like these types of functions that just mutate an argument and return essentially `void`. Pure functions with immutable data would be cool and nice to test (if/when we get to testing). But is striving for pure methods even that common in go? Is the power of pointers just too good to pass on than dealing with immutability?

pkg/platform/microservice/service.go
```diff
@@ -113,12 +108,99 @@ func (s *service) Create(responseWriter http.ResponseWriter, request *http.Reque
...
+				applicationID,
+				environment,
+			),
+		)
+		return
+	}
+
+	switch microserviceBase.Kind {
+	case platform.MicroserviceKindPurchaseOrderAPI:
+		// This is not ideal, we need to
```
joelhoisko, line 176:
```suggestion
		// TODO handle other updation operations too
```

Forgot to finish the sentence, here's my attempt at it

joelhoisko (changes requested):
- [x] Test that deleting and re-creating updates rawdatalog webhook
- [x] Test that updating purchase order api updates rawdatalog webhook
  - Tested it out with this [example json](https://gist.github.com/joelhoisko/03348885728dd170e605841817ab8fd0). The `username:password` combo is set to `joel:joel` in base64. I sent it with this command:
```shell
cat poapi.json | curl -X PUT -H "x-shared-secret: FAKE" -H "Tenant-ID: 453e04a7-4f9d-42f2-b36c-d51fa2c83fa3" -H "User-ID: local-dev" -H "Content-Type: application/json" localhost:8081/microservice -d @-
```

Otherwise nice stuff, found a few things to maybe talk about.

pkg/platform/microservice/rawdatalog/repo.go
woksin: I agree. Let's return the modified objects!

pkg/platform/microservice/service.go
woksin: Opps :D

pkg/platform/microservice/purchaseorderapi/handler.go
woksin: The function names are a bit awkward as they do similar, opposite things, but just a little bit differently

woksin: I don't support adding the boolean, that's not the "clean" way 😄 Mr. Bob does not agree https://twitter.com/unclebobmartin/status/1114137614377005057?lang=en

woksin: However I'll have a second look at the functions to see if I can come up with some new names

joelhoisko (approved):
👍

PR #50, opened 2021-11-04
Endpoints for some credentials

freshteapot:
## Summary

- Endpoint for service-account credentials
- Endpoint for container-registry credentials

### DIFF
--- a/pkg/platform/cicd/service.go
+++ b/pkg/platform/cicd/service.go
@@ -9,7 +9,6 @@ import (
 	"github.com/dolittle-entropy/platform-api/pkg/utils"
 	"github.com/gorilla/mux"
 	"github.com/sirupsen/logrus"
-	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
 type service struct {
@@ -29,9 +28,9 @@ func (s *service) GetServiceAccountCredentials(w http.ResponseWriter, r *http.Re
 	vars := mux.Vars(r)
 	applicationID := vars["applicationID"]
 	userID := r.Header.Get("User-ID")
-	tenantID := r.Header.Get("Tenant-ID")
+	customerID := r.Header.Get("Tenant-ID")
 
-	allowed := s.k8sDolittleRepo.CanModifyApplicationWithResponse(w, tenantID, applicationID, userID)
+	allowed := s.k8sDolittleRepo.CanModifyApplicationWithResponse(w, customerID, applicationID, userID)
 	if !allowed {
 		return
 	}
@@ -44,25 +43,14 @@ func (s *service) GetServiceAccountCredentials(w http.ResponseWriter, r *http.Re
 	logContext.Info("requested credentials")
 
 	secretName := "azure-devops"
-	secret, err := s.k8sDolittleRepo.GetSecret(applicationID, secretName)
+	secret, err := s.k8sDolittleRepo.GetSecret(logContext, applicationID, secretName)
 	if err != nil {
-		if !k8serrors.IsNotFound(err) {
-			logContext.WithFields(logrus.Fields{
-				"error":      err,
-				"secretName": secretName,
-			}).Error("issue talking to cluster")
-
-			utils.RespondWithError(w, http.StatusInternalServerError, "Something went wrong")
+		if err == platform.SecretNotFound {
+			utils.RespondWithError(w, http.StatusNotFound, fmt.Sprintf("Secret %s not found in application %s", secretName, applicationID))
 			return
 		}
 
-		logContext.WithFields(logrus.Fields{
-			"error":      err,
-			"secretName": secretName,
-			"hint":       "azure-devops-missing",
-		}).Error("application")
-
-		utils.RespondWithError(w, http.StatusNotFound, fmt.Sprintf("Secret %s not found in application %s", secretName, applicationID))
+		utils.RespondWithError(w, http.StatusInternalServerError, "Something went wrong")
 		return
 	}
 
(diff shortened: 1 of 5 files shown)


freshteapot:
# What
- Able to get the service-account credentials to be able to access the cluster with the same rights as a developer
- Able to get the container-registry credentials

pkg/platform/cicd/service.go
```diff
@@ -0,0 +1,116 @@
...
+
+			utils.RespondWithError(w, http.StatusInternalServerError, "Something went wrong")
+			return
+		}
+
+		logContext.WithFields(logrus.Fields{
+			"error":      err,
+			"secretName": secretName,
+			"hint":       "azure-devops-missing",
+		}).Error("application")
```
joelhoisko, line 63: This is something that I've been wondering for a while now and it might just be me not being familiar with structured logging or logging in go overall.

But I'm wondering why is the error message `application` and why is there a `hint` property (that is describing the error like the error message should?) instead of just writing out a descriptive error message like:
```suggestion
		logContext.WithFields(logrus.Fields{
			"error":      err,
			"secretName": secretName,
		}).Error("couldn't find the azure-devops secret")
```

I see this elsewhere in the codebase too (really short and non-descriptive error messages) and I'm always a bit confused about what are our logging practices in platform-api.

Should we start using a `hint` property instead of the `msg` property of the error message to understand it? To me the error message should already be the "hint" of the error (alongside with all of the other nice properties we put into the structured logs).

pkg/platform/cicd/service.go
```diff
@@ -0,0 +1,116 @@
...
+
+	allowed := s.k8sDolittleRepo.CanModifyApplicationWithResponse(w, tenantID, applicationID, userID)
+	if !allowed {
+		return
+	}
+
+	logContext := s.logContext.WithFields(logrus.Fields{
+		"method":      "GetServiceAccountCredentials",
+		"credentials": "serviceAccount",
+		"userID":      userID,
```
joelhoisko, line 42:
```suggestion
		"userID":      userID,
		"applicationID": applicationID,
		"customerID": tenantID,
```

More values in the logs the merrier right?

pkg/platform/cicd/service.go
```diff
@@ -0,0 +1,116 @@
...
+
+	allowed := s.k8sDolittleRepo.CanModifyApplicationWithResponse(w, tenantID, applicationID, userID)
+	if !allowed {
+		return
+	}
+
+	logContext := s.logContext.WithFields(logrus.Fields{
+		"method":      "GetServiceAccountCredentials",
+		"credentials": "containerRegistry",
+		"userID":      userID,
```
joelhoisko, line 87:
```suggestion
		"userID":      userID,
		"applicationID": applicationID,
		"customerID": tenantID,
```

Same here.

pkg/platform/cicd/service.go
```diff
@@ -0,0 +1,116 @@
...
+
+			utils.RespondWithError(w, http.StatusInternalServerError, "Something went wrong")
+			return
+		}
+
+		logContext.WithFields(logrus.Fields{
+			"error":      err,
+			"secretName": secretName,
+			"hint":       "acr-missing",
+		}).Error("application")
```
joelhoisko, line 108: Same comment here about the log messages.

pkg/platform/cicd/service.go
```diff
@@ -0,0 +1,116 @@
...
+
+func NewService(logContext logrus.FieldLogger, k8sDolittleRepo platform.K8sRepo) *service {
+	s := &service{
+		logContext:      logContext,
+		k8sDolittleRepo: k8sDolittleRepo,
+	}
+	return s
+}
+
+func (s *service) GetServiceAccountCredentials(w http.ResponseWriter, r *http.Request) {
```
joelhoisko, line 28: Looking at this method and comparing it to the `GetContainerRegistryCredentials` they are almost identical. Not sure if this is an anti-go thing to say but you could maybe make a more generic private function that would take the `secretName` in as a parameter (alongside the ID's etc). What do you think?

pkg/platform/cicd/service.go
```diff
@@ -0,0 +1,116 @@
+package cicd
```
joelhoisko, line 1: What does `cicd` mean? **C**ustomer **I**dentified **C**re**D**entials? Could the name be something a bit more descriptive, maybe something like `credentials`?

joelhoisko (changes requested):

pkg/platform/cicd/service.go
freshteapot: Why cicd?

> In software engineering, CI/CD or CICD is the combined practices of continuous integration (CI) and either continuous delivery or continuous deployment (CD).[1] 

https://en.wikipedia.org/wiki/CI/CD

joelhoisko: Yeah I understood that when I started going through the Studio PR. I was just pronouncing it in my head as kikd and I couldn't understand it for the life of me what It meant (and as the package currently is about giving credentials so I assumed it was dealing with that).

pkg/platform/cicd/service.go
freshteapot: - I have removed hint.
- Moved the logContext into repo.GetSecrets
- Using a generic "secret not found"

pkg/platform/cicd/service.go
freshteapot: Done.

pkg/platform/cicd/service.go
freshteapot: Done

pkg/platform/cicd/service.go
freshteapot: - I have removed hint.
- Moved the logContext into repo.GetSecrets
- Using a generic "secret not found"

pkg/platform/cicd/service.go
freshteapot: Refactored a little.
Im hesitant to go further.
When we bring "github credentials" to the end user, we will have enough examples of how to refactor in a generic way.

pkg/platform/cicd/service.go
freshteapot: It could go under application as a sub-folder.
But so could many other parts.

It reminds me we should bubble up https://app.asana.com/0/1200181647276434/1201235587155608/f

joelhoisko (approved):
Nice 👍

PR #51, opened 2021-11-04
applications in /applications endpoint return empty []

freshteapot:
## Summary

Querying the /applications data, will now return an empty array, not null.


freshteapot (commented):
**Approved**

Based on

```sh
kubectl -n system-api \       
patch deployments.app dev-api-v1 \
--type='json' -p='[{"op": "replace", "path": "/spec/template/spec/containers/0/image", "value":"dolittle/platform-api:bug-applications-null"}]'
```

PR #52, opened 2021-11-05
Not json + first need to lookup service account

freshteapot:
## Summary
- Specific endpoint for devops to not expose all service accounts.
- Lookup service account to find the secret name.

### DIFF
--- a/pkg/platform/cicd/service.go
+++ b/pkg/platform/cicd/service.go
@@ -23,9 +23,9 @@ func NewService(logContext logrus.FieldLogger, k8sDolittleRepo platform.K8sRepo)
 	return s
 }
 
-func (s *service) GetAzureDevops(w http.ResponseWriter, r *http.Request) {
+func (s *service) GetDevops(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
-	vars["name"] = "azure-devops"
+	vars["name"] = "devops"
 	mux.SetURLVars(r, vars)
 	s.getServiceAccountCredentials(w, r)
 }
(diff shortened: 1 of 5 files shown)


freshteapot:
Follow up from #50.

pkg/platform/cicd/service.go
```diff
@@ -24,9 +23,18 @@ func NewService(logContext logrus.FieldLogger, k8sDolittleRepo platform.K8sRepo)
 	return s
 }
 
-func (s *service) GetServiceAccountCredentials(w http.ResponseWriter, r *http.Request) {
+func (s *service) GetAzureDevops(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	vars["name"] = "azure-devops"
```
joelhoisko, line 28: Currently the only reason why we use the words "Azure" in here is because we decided to when we created the service account to name it like that. But in the future this service account would be used for all sorts of other CI/CD pipelines, scripts etc. There is nothing specific to Azure Devops about the service account.

So maybe having something more generic would be nice, maybe just a simple "devops" (as loaded of a term as it is, I don't really like it but you get the idea). This would also mean that we'd have to rename the service account in operations and break somebodys workflow again so it might not be the best right now so I understand if we just wanna get this done now. But in the future I would much prefer that this service account would be more generic, or that the user can create specific service accounts with specific permissions for their different CI/CD needs (which is a lot more work).

joelhoisko (changes requested):

joelhoisko (approved):

PR #61, opened 2021-11-15
Add ability to create a base microservice without a Runtime container

joelhoisko:
## Summary

Add's the ability to create a base microservice without a Runtime container by defining the wanted Runtime image as `"none"` in the request.

Also adds memory limits to newly created Runtime containers in a microservice. The request is for 250MB and the limit is 1GB.

Related to https://github.com/dolittle/Studio/pull/136


freshteapot (approved):
Nice work :)
Works locally.

PR #65, opened 2021-11-29
Public url for microservices in the platform but not in studio

freshteapot:
## Summary

- Rename HttpResponseApplications2 to HttpResponseApplication
- Expose the Ingress host + path per customer Tenant ID
- Expose Unique Ingress Paths (microservice paths)

### DIFF
--- a/pkg/platform/k8s_repo.go
+++ b/pkg/platform/k8s_repo.go
@@ -3,7 +3,6 @@ package platform
 import (
 	"bytes"
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -736,10 +735,8 @@ func (r *K8sRepo) GetIngressHTTPIngressPath(applicationID string, environment st
 				// For now, there is no need to expose it
 				rulePath.Backend = networkingv1.IngressBackend{}
 
-				diffA, _ := json.Marshal(rulePath)
 				exists := funk.Contains(items, func(item networkingv1.HTTPIngressPath) bool {
-					diffB, _ := json.Marshal(item)
-					return string(diffA) == string(diffB)
+					return rulePath == item
 				})
 
 				if exists {
--- a/pkg/platform/k8s_repo.go
+++ b/pkg/platform/k8s_repo.go
@@ -3,10 +3,12 @@ package platform
 import (
 	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"net/http"
+	"regexp"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


freshteapot:
# Info
- Linked to https://github.com/dolittle/Studio/pull/139

pkg/platform/k8s_repo.go
```diff
@@ -653,3 +670,94 @@ func (r *K8sRepo) createAnnotationsAndLabels(customerID, customerName, applicati
...
+		for _, rule := range ingress.Spec.Rules {
+			for _, rulePath := range rule.HTTP.Paths {
+				// For now, there is no need to expose it
+				rulePath.Backend = networkingv1.IngressBackend{}
+
+				diffA, _ := json.Marshal(rulePath)
+				exists := funk.Contains(items, func(item networkingv1.HTTPIngressPath) bool {
+					diffB, _ := json.Marshal(item)
+					return string(diffA) == string(diffB)
+				})
```
joelhoisko, line 743: This led me down the rabbit hole on how to compare structs in golang and I found that structs are comparable with the basic [`==` equality operator](https://www.geeksforgeeks.org/structure-equality-in-golang/). So is there a reason you're going the way of JSONifying the structs first? Like relying on some of the structs JSON annotations to clean up the strings?

joelhoisko (changes requested):
Nice work and tests too :) One question about comparing structs though.

pkg/platform/k8s_repo.go
freshteapot: I can't explain it, with proof. So we shall use the "==", its cleaner :).
If I ever stumble upon the reason why, I will come back to you. But for now, I am keen to embrace this as its cleaner.

I thought there was something about pointers, but the documentation tells me I am wrong and running the code in question with some printing confirmed, whatever I am trying to sell, does not apply here :P.

```go
for _, rule := range ingress.Spec.Rules {
	for _, rulePath := range rule.HTTP.Paths {
		// For now, there is no need to expose it
		rulePath.Backend = networkingv1.IngressBackend{}

		diffA, _ := json.Marshal(rulePath)
		fmt.Println("HELLO", len(items))
		items = append(items, rulePath)
		exists := funk.Contains(items, func(item networkingv1.HTTPIngressPath) bool {
			diffB, _ := json.Marshal(item)
			fmt.Println("operator match", rulePath == item)
			fmt.Println("string match", string(diffA) == string(diffB))
			return string(diffA) == string(diffB)
		})

		if exists {
			continue
		}
		items = append(items, rulePath)
	}
}
```

joelhoisko (approved):

PR #66, opened 2021-12-01
WIP: add user to customer

freshteapot:
## Summary

Summary of the PR here. The GitHub release description is created from this comment so keep it nice and descriptive.

Remember to remove sections that you don't need or use.

### Added

- Describe the added features

### Changed

### Fixed

### Removed

- Describe what was removed and why

### Security

### Deprecated


cmd/tools/users/get-user.go
```diff
@@ -0,0 +1,53 @@
...
+			fmt.Println("An email is required")
+			return
+		}
+
+		customerID, _ := cmd.Flags().GetString("customer-id")
+		if customerID == "" {
+			fmt.Println("A customerID is required")
+			return
+		}
+
```
joelhoisko, line 31:
```suggestion
```

The `customerID` variable isn't used in the code anywhere so it can be deleted

cmd/tools/users/root.go
```diff
@@ -0,0 +1,18 @@
...
+	Use:   "users",
+	Short: "Platform tools for users",
+	Long: `
+
+Tools to interact with the users in the platform`,
+}
+
+func init() {
+	RootCMD.AddCommand(addUserToCustomerCMD)
+	RootCMD.AddCommand(listEmailsCMD)
```
joelhoisko, line 17:
```suggestion
	RootCMD.AddCommand(listEmailsCMD)
	RootCMD.AddCommand(getUserCMD)
```

The `get-user` command wasn't available as it wasn't part of the root command.

joelhoisko (changes requested):
Tried it out and it worked fine. I was expecting the `add-user-command` to actually add the user to Kratos instead of just creating and showing me the command on how to do it. But it worked so this does make life easier.

cmd/tools/users/get-user.go
freshteapot: Done

freshteapot:
> Tried it out and it worked fine. I was expecting the `add-user-command` to actually add the user to Kratos instead of just creating and showing me the command on how to do it. But it worked so this does make life easier.

You are right, I have added a few todo's, it makes sense.
I also realise, that we need to figure out later how to also add the user to the correct group in azure.

With the work on application-namespace branch, we could rely on terraform to do the heavy lifting or we learn how todo it the golang way. In 2022. We will figure it out :).

joelhoisko (approved):
Nice to see this talking directly to Kratos now 👍

freshteapot:
# Todo
- [x] Make config of where kratos is configurable

# How
- Setup port-forward
```sh
kill -9 $(lsof -ti tcp:4434)
kubectl -n system-authentication port-forward $(kubectl get pod -l "component=Users,environment=Dev" -o name -n system-authentication) 4434:4434 &
```
- Add customer to user by email
```sh
go run main.go tools users  add customer --email="xxx@dolittle.com" --customer-id="XXX"
```

- Add customer to user by userID
```sh
go run main.go tools users  add customer --user-idl="XXX" --customer-id="XXX"
```


# Reference
- https://app.asana.com/0/1201338547604691/1201451644838461/f

PR #68, opened 2021-12-15
Adding ModuleName + applicationID

freshteapot:
## Summary

- Building the application info to include ModuleName and ApplicationID.
- Terraform create customer now includes the module_name.


freshteapot:
# Reference
- https://app.asana.com/0/1201408587836627/1201519691135424/f

PR #69, opened 2021-12-15
Add restart method and make signatures a little similar

freshteapot:
## Summary

- Api to restart microservice


freshteapot:
# Reference
- https://app.asana.com/0/1201408587836627/1201437467283837/f
- Run with https://github.com/dolittle/Studio/pull/142

PR #72, opened 2022-01-14
Add script for adding tenants to resources.json

joelhoisko:
## Summary

Add's a simple script for reading a `microservice.yml` file, adding new tenants to the `resource.json` and outputting the `resource.json` into the terminal for copy-pastining it back into the `microservice.yml` file.

### DIFF
--- a/cmd/tools/add-tenants-to-resources.go
+++ b/cmd/tools/add-tenants-to-resources.go
@@ -1,0 +1,164 @@
+package tools
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+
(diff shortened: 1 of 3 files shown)


joelhoisko:
I don't feel like this code is really that useful because as we're going to be [splitting the configmaps](https://github.com/dolittle/platform-api/pull/73) to their own files it is a lot easier to do changes like this through k8s sdk and then just pull the changed configmaps. Closing it now as well as the [Asana ticket.](https://app.asana.com/0/1201643711163095/1201631150914718)

PR #73, opened 2022-01-19
Add commands to pull configmaps & deployments and to update them with the `platform.json`

joelhoisko:
## Summary

- Adds new command `pull-dolittle-config` which pulls all of the `*-dolittle` configmaps from the cluster and writes them into the `Source/V3/Kubernetes/Customers/<customer>/<application>/<environment>/<microservice>/microservice-configmap-dolittle.yml` files. Related PR https://github.com/dolittle-platform/Operations/pull/150
Usage:
```sh
go run main.go tools automate pull-dolittle-config <repo-root>
```

- Adds new command `add-platform-config` which updates the `*-dolittle` configmaps in the cluster with the new `platform.json` resource and also updates the microservices deployments Runtime container to have a volumeMount for the `platform.json`
Usage:
```sh
	go run main.go tools automate add-platform-config \
	--application-id="11b6cf47-5d9f-438f-8116-0d9828654657" \
	--environment="Dev" \
	--microservice-id="ec6a1a81-ed83-bb42-b82b-5e8bedc3cbc6" \
	--dry-run=true
```

- Adds new command `get-microservices-metadata` which outputs a JSON array of all of the metadata of all of the clusters microservices

- Adds new command `pull-microservice-deployment` which pulls all microservices deployments from the cluster and writes them into `Source/V3/Kubernetes/Customers/<customer>/<application>/<environment>/<microservice>/microservice-deployment.yml` files. Related PR https://github.com/dolittle-platform/Operations/pull/155
Usage:
```sh
go run main.go tools automate pull-microservice-deployment <repo-root>
```

- Adds new command `import-dolittle-configmaps` which ca n be used to bootstrap your local cluster. From PR https://github.com/dolittle/platform-api/pull/74

- When pulling the deployments or configmaps and converting the given resource to a `configK8s.Microservice` struct, if the given resource doesn't have a `dolittle.io/microservice-kind` specified it will default to the new `MicroserviceKindUnknown`. This is only internal for now and it's not set to the k8s resources annotations.
- Adds new package `automate` to help with the `platform tools automate` commands


joelhoisko:
We should also discuss if the `managedFields` should be removed from the output.

pkg/platform/automate/share.go
```diff
@@ -0,0 +1,269 @@
...
+}
+
+func UpdateConfigMap(ctx context.Context, client kubernetes.Interface, logContext logrus.FieldLogger, configMap corev1.ConfigMap, dryRun bool) error {
+	microservice := ConvertObjectMetaToMicroservice(configMap.GetObjectMeta())
+	platform := configK8s.NewMicroserviceConfigmapPlatformData(microservice)
+	b, _ := json.MarshalIndent(platform, "", "  ")
+	platformJSON := string(b)
+
+	if configMap.Data == nil {
+		// TODO this is a sign it might not be using a runtime, maybe we skip
```
freshteapot, line 238: Should we warn?
I cant decide if we should return an error here.
Perhaps later, when we do know the microservice kind, we can add logic to handle if it should do what it currently does, or return an error.

Ie simple == includes runtime and should error

Hmm I just realised, we can have "simple" and not use the runtime.... we might have to return to this.

freshteapot (changes requested):
Almost there :)

cmd/tools/automate/update-dolittle-config.go
```diff
@@ -0,0 +1,183 @@
...
+				application := namespace.Labels["application"]
+				logContext = logContext.WithFields(logrus.Fields{
+					"customer":    customer,
+					"application": application,
+				})
+
+				configMaps, err := automate.GetDolittleConfigMaps(ctx, client, namespace.GetName())
+
+				if err != nil {
+					// TODO this might be to strict, perhaps we have a flag to let them skip?
```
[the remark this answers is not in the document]
joelhoisko: Looking at the code the only way the `automate.GetDolittleConfigMaps()` is going to error is if the underlying `client.CoreV1().ConfigMaps(namespace).List()` call is going to error. I don't think that call would error if the namespace would have no ConfigMaps so the only error then would be a problem with k8s which would grind everything to a halt anyways.

But if a namespace can fail independently then it would make sense to be able to just have a warning for it. I think we can let it be like this for now and if it becomes a problem we can add something like `--skip-errors` flag.

cmd/tools/automate/update-dolittle-config.go
```diff
@@ -0,0 +1,183 @@
...
+func processOne(ctx context.Context, client kubernetes.Interface, logContext logrus.FieldLogger, dryRun bool, applicationID string, environment string, microserviceID string) {
+	logContext = logContext.WithFields(logrus.Fields{
+		"application_id":   applicationID,
+		"environment":      environment,
+		"microservivce_id": microserviceID,
+	})
+
+	configMap, err := automate.GetOneDolittleConfigMap(ctx, client, applicationID, environment, microserviceID)
+	if err != nil {
+		// TODO this might be to strict, perhaps we have a flag to let them skip?
```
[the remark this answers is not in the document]
joelhoisko: This is in the `processOne()` function, and if the function can't get the configmap that it should process it makes sense that the function errors as it can't continue to the next step, which would be to update the not-found configmap meaning the function breaks.

The function could return return a `(err error, ok bool)` which we could check that did it do it's job correctly though. Then we could have the same kind of `--skip-errors` flag from my other comment.

pkg/platform/automate/share.go
joelhoisko: I started to think first that I don't think it's the `UpdateConfigMap()` functions job to check and make assumptions about what kind of configmap it's supposed to update. But then I realised that this function doesn't actually just update configmaps as a generic thing, but rather only adds the specific `platform.json` part into it. In that case either this function (which could maybe be called `AddMicroserviceConfigmapPlatformData()` or something equally ugly and verbose) or it's caller should make sure that the given configmap is a dolittle configmap.

Currently I don't think adding some new data to the dolittle configmaps (even the ones that don't use the Runtime) would create any problems except it would add noise to the configmaps. But you could make the same argument for adding the `platform.json` data into microservices that aren't going to use it...

And as you said, we don't really use the `microservice-kind` annotation in a defined way currently so I don't think you can draw too many conclusions from it.

So in the end I'm fine if we error there or write a warning and skip the update.

joelhoisko:
One thing that I also learned just now is that in platform-api when writing fields for the logger we use both `microservice_id` and `microserviceID` fields. Which one should be our standard? From a JSON perspective I like the camelCased version more as it feels more JSONy (you wouldn't write JS objects with snake_cased properties).

cmd/tools/automate/update-dolittle-config.go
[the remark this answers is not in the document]
joelhoisko: Or maybe `--ignore-errors` sounds more professional

pkg/platform/automate/share.go
```diff
@@ -0,0 +1,269 @@
...
+
+	environment := labels["environment"]
+
+	return configK8s.Microservice{
+		ID:          microserviceID,
+		Name:        microserviceName,
+		Tenant:      customerTenant,
+		Application: k8sApplication,
+		Environment: environment,
+		// TODO wwhen we explicitly set the kind as an annotation we can rely on it
```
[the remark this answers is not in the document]
freshteapot: If we dont have ["dolittle.io/microservice-kind"] default to unknown.

freshteapot:
> One thing that I also learned just now is that in platform-api when writing fields for the logger we use both `microservice_id` and `microserviceID` fields. Which one should be our standard? From a JSON perspective I like the camelCased version more as it feels more JSONy (you wouldn't write JS objects with snake_cased properties).

I created a ticket to reflect your concerns, I am not sold on the "js object" reason :).
I do not hold a strong opinion on either way.

You are very welcome to pull the task into the iteration and make the changes you want, after this PR lands?

https://app.asana.com/0/1201325052247030/1201713848182574/f

pkg/platform/automate/share.go
freshteapot: We should rename this function to make it clearer what it is doing.

microservice-kind, is going to be added https://app.asana.com/0/1201325052247030/1201451644838466/f.

cmd/tools/automate/update-dolittle-config.go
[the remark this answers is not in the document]
freshteapot: I think we should wait before skipping errors logic.
We might only run the script once.

Just making the script has led to progress on how to make the next script easier :).

cmd/tools/automate/update-dolittle-config.go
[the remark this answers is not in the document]
freshteapot: We should find out :).
I think if the error is not found we should move on.
If the error is anything but, it should fail.

Let us start by taking the errors seriously :).

pkg/platform/automate/share.go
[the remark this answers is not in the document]
joelhoisko: I'm defaulting to the new `MicroserviceKindUnknown` in this function now if the kind isn't specified in the resource. It's a step at least.

pkg/platform/automate/share.go
joelhoisko: I've renamed it to `AddDataToConfigMap()`

joelhoisko:
Copying the todo from the top level summary here before I forget:

# Todo
- [x] We should also discuss if the managedFields should be removed from the output.
- [ ] ~~Test in local cluster~~ (not feasible with the deployments as I can't just write a script to import the ones from the cluster)
- [x] Hook up platform.json to volume
- [x] ConvertObjectMetaToMicroservice check for microservice-kind, default to unknown

joelhoisko:
# Dry run test

1. Change your context to the prod cluster
2. Run this command to only print out the changed configmap and then the deployment (it's dry run):
```sh
go run main.go tools automate add-platform-config \
        --application-id="11b6cf47-5d9f-438f-8116-0d9828654657" \
        --environment="Dev" \
        --microservice-id="ec6a1a81-ed83-bb42-b82b-5e8bedc3cbc6" \
        --dry-run=true | jq -r '.data'
```
3. Look at the data in the terminal and see that the `platform.json` volumeMount is added to the runtime container in the deployment and that the configmap also has a `platform.json` key added to it.

cmd/tools/automate/pull-microservice-deployment.go
```diff
@@ -0,0 +1,125 @@
...
+				"customer":    customer,
+				"application": application,
+			})
+
+			deployments, err := automate.GetDeployments(ctx, client, namespace.GetName())
+			if err != nil {
+				logContext.Fatal("Failed to get deployments")
+			}
+
+			// var runtimeDeployments []appsv1.Deployment
```
freshteapot, line 77: Please remove the comments here.

cmd/tools/automate/pull-dolittle-config.go
```diff
@@ -0,0 +1,113 @@
...
+
+			logContext.WithFields(logrus.Fields{
+				"totalConfigMaps": len(configMaps),
+			}).Info("Found dolittle configmaps")
+
+			if dryRun {
+				continue
+			}
+
+			err = writeConfigMapsToDirectory(args[0], configMaps, scheme, serializer)
```
freshteapot, line 85: What is the downside of moving writeConfigMapsToDirectory into pkg/automate as WriteConfigMapsToDirectory?

cmd/tools/automate/pull-microservice-deployment.go
```diff
@@ -0,0 +1,125 @@
...
+
+			logContext.WithFields(logrus.Fields{
+				"totalDeployments": len(deployments),
+			}).Info("Found microservice deployments")
+
+			if dryRun {
+				continue
+			}
+
+			err = writeDeploymentsToDirectory(args[0], deployments, scheme, serializer)
```
freshteapot, line 95: What is the downside of moving writeDeploymentsToDirectory into pkg/automate as WriteDeploymentsToDirectory?

pkg/platform/automate/command_input.go
```diff
@@ -0,0 +1,14 @@
+package automate
+
+import (
+	"encoding/json"
+
+	"github.com/dolittle/platform-api/pkg/platform"
+)
+
+// ParseMicroserviceMetadata Given a json oneliner, return the microservice metadata
+func ParseMicroserviceMetadata(input string) (platform.MicroserviceMetadataShortInfo, error) {
```
freshteapot, line 10: Better name :)

pkg/platform/automate/share.go
```diff
@@ -0,0 +1,390 @@
...
+
+func InitializeSchemeAndSerializerForConfigMap() (*runtime.Scheme, *k8sJson.Serializer, error) {
+	// based of https://github.com/kubernetes/kubernetes/issues/3030#issuecomment-700099699
+	// create the scheme and make it aware of the corev1 & appv1 types
+	scheme := runtime.NewScheme()
+	err := corev1.AddToScheme(scheme)
+	if err != nil {
+		return scheme, nil, err
+	}
+	err = appsv1.AddToScheme(scheme)
```
freshteapot, line 69: This no longer reflects the func signature.
Is there a downside to adding many schemes?
If not maybe we rename the  the func.

Otherwise we might need to copy and paste for each resource.

pkg/platform/automate/share.go
```diff
@@ -0,0 +1,390 @@
...
+		if _, ok := deployment.Annotations["dolittle.io/microservice-id"]; !ok {
+			continue
+		}
+		microserviceDeployments = append(microserviceDeployments, deployment)
+	}
+	return microserviceDeployments, nil
+}
+
+// GetRuntimeDeployment get's a microservices deployment and the index of the Runtime container within that deployment
+func GetRuntimeDeployment(ctx context.Context, client kubernetes.Interface, applicationID string, environment string, microserviceID string) (int, appsv1.Deployment, error) {
```
freshteapot, line 165: I think we should refactor this function out.
We have K8sGetDeployment
```go
deployment, err := microserviceK8s.K8sGetDeployment(r.k8sClient, ctx, namespace, microserviceID)
```

Let us move the func here.
We can add a note to come back to this file after we land applications. To tidy up some of the structure.

I would like to reorder the context and client in K8sGetDeployment.
Then use this function to lookup the microservice.
Then a helper a function that takes a deployment and returns the "index, container".

```go
func GetContainerIndex(deployment appsv1.Deployment, name string) int {
	for index, container := range deployment.Spec.Template.Spec.Containers {
		if container.Name == name {
			return index
		}
	}
	return -1
}
```

pkg/platform/automate/share.go
```diff
@@ -0,0 +1,390 @@
...
+	}
+
+	for _, deployment := range deployments {
+		microservice := ConvertObjectMetaToMicroservice(deployment.GetObjectMeta())
+		microservices = append(microservices, microservice)
+	}
+	return microservices, nil
+}
+
+func GetNamespaces(ctx context.Context, client kubernetes.Interface) []corev1.Namespace {
```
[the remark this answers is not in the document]
freshteapot: Can we replace the use of getNamespaces in tools/studio to use this and then delete the function.

pkg/platform/automate/share.go
```diff
@@ -0,0 +1,390 @@
...
+	namespacesList, err := client.CoreV1().Namespaces().List(ctx, metav1.ListOptions{})
+	if err != nil {
+		panic(err.Error())
+	}
+	return namespacesList.Items
+}
+
+func IsApplicationNamespace(namespace corev1.Namespace) bool {
```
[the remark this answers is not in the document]
freshteapot: Can we replace the use of isApplicationNamespace in tools/studio to use this and delete the function.

pkg/platform/automate/share.go
```diff
@@ -0,0 +1,390 @@
...
+
+	if dryRun {
+		bytes := SerializeRuntimeObject(&deployment)
+		logContext.WithFields(logrus.Fields{
+			"data": string(bytes),
+		}).Info("--dry-run enabled, skipping the update")
+		return nil
+	}
+
+	_, err := client.AppsV1().Deployments(deployment.Namespace).Update(ctx, &deployment, metav1.UpdateOptions{})
```
freshteapot, line 384: Love it!

pkg/platform/automate/share.go
```diff
@@ -0,0 +1,390 @@
...
+
+	environment := labels["environment"]
+
+	var kind platform.MicroserviceKind
+	if kindString, ok := annotations["dolittle.io/microservice-kind"]; !ok {
+		kind = platform.MicroserviceKindUnknown
+	} else {
+		kind = platform.MicroserviceKind(kindString)
+	}
+
```
freshteapot, line 215: The use of else is not needed here. I guess it might be a style thing, but we have a default "unknown" and we plan on changing it "if" something is different.

We also have very similar logic in https://github.com/dolittle/platform-api/blob/c978b7ba019986fe33a3077064facb14dc6c1e31/pkg/platform/k8s_repo.go#L199-L202

Which makes me believe we should hoist this into a func, and then clean up the code in both places.
To start with, I think the signature could be "annotations". and return kind, if its finds it, or unknown.


```suggestion
	kind := platform.MicroserviceKindUnknown
	if kindString, ok := annotations["dolittle.io/microservice-kind"]; ok {
		kind = platform.MicroserviceKind(kindString)
	}```

freshteapot (changes requested):
We are getting there :)

pkg/platform/automate/share.go
joelhoisko: Note for myself: move things to automate

pkg/platform/automate/share.go
joelhoisko: I just changed the function name to `InitializeSchemeAndSerializer()`. I don't know if there's any downside to adding many schemes, but I assume it's negligible. Digging behind the scenes the `AddToScheme()` call essentially just adds some more stuff to the schemes `gvkToType map[schema.GroupVersionKind]reflect.Type` and `typeToGVK map[reflect.Type][]schema.GroupVersionKind` properties. And it just adds tens of elements so I think we're all fine here.

pkg/platform/automate/share.go
joelhoisko: - I removed the `K8sGetDeployment()` function
- I added `GetDeployment()` to automate that takes a namespace and microserviceID, getting rid of the applicationID and environment
- I added a `GetContainerIndex()` function to automate
- I removed the `GetRuntimeDeployment()` function
- Fixed all the places that used the removed functions to use the new ones

pkg/platform/automate/share.go
joelhoisko: Yeah I was debating myself that should I be setting the variable first by default to unknown or using an if/else.

I created function `GetMicroserviceKindFromAnnotations` in the platform package. I couldn't make it inside `automate` as we'd get a cyclical import (`automate` imported the `platform.MicroserviceKind` type, `platform` imported `automate` to get the kind -> cyclical import).

cmd/tools/automate/pull-dolittle-config.go
joelhoisko: I was thinking that the pkg/automate was for generic stuff that can be shared, but I guess the `WriteConfigMapsToDirectory()` function could also be used by other pieces of code someday. I moved it and removed the `scheme` and `serializer` from the args as it doesn't make sense to have to pass it in there anymore.

cmd/tools/automate/pull-microservice-deployment.go
joelhoisko: I moved it, see my comment about `WriteConfigMapsToDirectory()`

pkg/platform/automate/share.go
freshteapot: Nice :) I am excited by this func.

cmd/tools/automate/pull-dolittle-config.go
freshteapot: okay :).

cmd/tools/automate/pull-microservice-deployment.go
freshteapot: Noted, we will make the pkg structure better in due course.

pkg/platform/automate/share.go
freshteapot: Game on :D.

pkg/platform/automate/share.go
freshteapot: <3

freshteapot (approved):
Some nice features landing via this PR:

- Now we know how to dump beautiful yaml files (configmaps + deployments)
- Reading in ndjson to do things
- Able to update configmaps easier now (thinking of new tenants)
- We have a little recipe on how to update "things in the cluster".

Happy days

As I mentioned in teams.
I think when it comes to deploying this, we can hit all none production now.
Then we need to think a little how to handle production, as the update does trigger a restrart on the pods.

PR #75, opened 2022-01-21
Make release label action to only run on PRs that target main

joelhoisko:
## Summary

Make release label action to only run on PRs that target main. This is because it doesn't make sense to make versioned releases against a PR image as they aren't versioned (their tags are the github commit hash instead of a version).

Also adds a new `make server` command to the Makefile to make it easier to start the server without having to remember all of the envs you have to set.

Tested on https://github.com/dolittle/platform-api/pull/76


freshteapot (approved):
Nice

PR #77, opened 2022-01-21
Fix getServiceAccountCredentials to return the whole k8s Secret resource

joelhoisko:
## Summary

Fixes the `getServiceAccountCredentials()` method to return the whole k8s Secret resource. This is because Azure DevOps requires the whole Secret resource instead of just the data part when setting up a kubernetes connection.

The Studio button that calls this endpoint:
image

### DIFF
--- a/pkg/platform/cicd/service.go
+++ b/pkg/platform/cicd/service.go
@@ -70,7 +70,7 @@ func (s *service) getServiceAccountCredentials(w http.ResponseWriter, r *http.Re
 		return
 	}
 
-	utils.RespondWithJSON(w, http.StatusOK, secret.Data)
+	utils.RespondWithJSON(w, http.StatusOK, secret)
 }
 
 func (s *service) GetContainerRegistryCredentials(w http.ResponseWriter, r *http.Request) {


joelhoisko:
# Test

1. Setup the local k3d and change context to it
2. Give Customer-Chris a `devops` serviceaccount with our own `create-service-account` command as the k3d cluster doesn't have it by default (potential future fix):
```sh
go run main.go microservice create-service-account --all
```
3. Start up platform-api:
```sh
GIT_REPO_DIRECTORY="/tmp/dolittle-local-dev" \      
GIT_REPO_DIRECTORY_ONLY="true" \
GIT_REPO_DRY_RUN="true" \
GIT_REPO_BRANCH=main \
LISTEN_ON="localhost:8081" \
HEADER_SECRET="FAKE" \
AZURE_SUBSCRIPTION_ID="e7220048-8a2c-4537-994b-6f9b320692d7" \
go run main.go microservice server
```
4. Start up Studio frontend & backend
```sh
make develop-frontend
```
```sh
make develop-backend
```
5. Go to http://localhost:9007/selfservice/documentation/application/11b6cf47-5d9f-438f-8116-0d9828654657/Dev/ci-cd/azure-pipelines and copy the platform-to-pipeline credentials and see that it returns the whole k8s resource now.

To reduce noise we could remove things from the k8s resource (the usual `managedFields` and `resourceVersion` stuff) but this output isn't meant to be viewed, only pasted to Azure DevOps so I think it's fine like this.

freshteapot (approved):
Works locally, it does include the managedfields.
If you want to test it in devops, do it. Or we wait new feedback.

From my part, I am happy.

PR #79, opened 2022-01-27
Make `GetDeployment` environment aware

joelhoisko:
## Summary

Changes the `GetDeployment` function to require `applicationID` & `environment` instead of just a `namespace`. This is because some microservices reuse the same microserviceID accross environment. This makes it so that you can reliably always get the correct deployment from the correct environment.


freshteapot (approved):
It makes sense.

Now we have a sweeper ticket to check for more.
https://app.asana.com/0/1201325052247030/1201729501717327/f

PR #80, opened 2022-01-27
Make add-platform-config command more robust

joelhoisko:
## Summary

Fixes the `add-platform-config` command to skip deployments that don't have a Runtime container.

Also skips if the microservice doesn't have a *-dolittle configmap.

Also checks first if the deployment the volumeMount would be added to has a corresponding volume specified, otherwise the deployment would fail to start. In testing with dry-run I didn't see that logmessage pop up but it should still be handled just to be sure.


freshteapot (approved):
Using the dry-run command against production.
It works well :).

PR #82, opened 2022-02-02
Adding environment to GetMicrosericeName

freshteapot:
## Summary

Including the environment name in the lookup for microservice name


freshteapot:
verified in dev with https://github.com/dolittle/Studio/pull/149.

PR #83, opened 2022-02-02
Reduce the information sent via personalised-info api for documentation

freshteapot:
## Summary

- Reducing the amount of information used for the personalised-info, to that which is currently used in Studio.
- `GetStudioInfo` function to reduce the boiler plate of getting requried info per request.


freshteapot (commented):
🚢 🇮🇹  works on dev with https://github.com/dolittle/Studio/pull/150.

PR #59, opened 2021-11-10
Application namespace

freshteapot:
# Summary
- Able to create a customer via the command line
- Able to create a customer via the rest api
- Able to create an application via the command line
- Able to create an application via the rest api
- Able to create a microservice
- Developer rbac out the box, supports querying own namespace
- Creating an application will make a namespace
- Add Acr secret to access the container registry
- Application Storage
- Application Role
- Application RoleBindings
- Environment Networkpolicy
- Environment tenants configmap
- Mongo (Service, Stateful, Cronjob)
- For local dev we create rolebindings for the local-dev user (Linked to Studio and k3d)

PR #90, opened 2022-02-22
Refactor CLI commands

joelhoisko:
## Summary

Changed the CLI commands:
- New root command `platform template`.
	- `delete` for creating templates that delete things platform wide. Replaces old `platform tools studio delete-*` commands:
		- `application`
		- `customer`
- Refactored `platform tools studio` command:
  - `upsert` as all of these commands are about upsert the existing `*.json` files.
    - `terraform`
    - `studio`
    - `application`
  - `create` for creating new resources
    - `service-account`
  - `get` for getting Studio specific resources
    - `customers`
- Refactor `platform tools job` command:
	- `status` get's the job status
	- `template` for creating the pre-filled k8s Jobs
		- `customer`
		- `application`
- Refactor `platform tools terraform` command:
	- `template` for creating the pre-filled Terraform HCL
		- `customer`
		- `application`
- Moved `platform tools automate get-microservice-metadata` command to `platform tools explore microservices`

## Ref
- [Brainstorming Miro board](https://miro.com/app/board/uXjVOKmmIm4=/)
- [Asana ticket](https://app.asana.com/0/home/1142702759924347/1201235587155608)


joelhoisko:
The `helpers` command could also be renamed to `templates` to try to hammer down the idea that these commands only output stuff into stdout for you to copy paste. Or maybe there's a better name, as `templates` sounds like it deals with... templates which these things kinda are but kinda aren't.

cmd/tools/helpers/kubernetes/application-job.go
```diff
@@ -15,13 +15,13 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 )
 
-var createCustomerApplicationCMD = &cobra.Command{
-	Use:   "create-customer-application",
-	Short: "Create job to make customer application",
+var applicationJobCMD = &cobra.Command{
```
freshteapot, line 18: Why do you consider these helpers?

I struggled with this sub-package, as "jobs" has a meaning in kubernetes.
For me, I used to love "helpers" and "utils", but then I took heart when I was reading in golang about avoid using "utils" for it doesnt provide context. I came to realise "helpers" also dont provide context.

https://github.com/golang/go/wiki/CodeReviewComments#package-names

Changing this and the customer one, also means you need to dig into the code and change files within https://github.com/dolittle/platform-api/tree/main/pkg/platform/job/k8s which call these functions.

cmd/tools/helpers/root.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/dolittle/platform-api/cmd/tools/helpers/kubernetes"
+	"github.com/dolittle/platform-api/cmd/tools/helpers/terraform"
+	"github.com/spf13/cobra"
+)
+
+var RootCMD = &cobra.Command{
+	Use:   "helpers",
+	Short: "Helpers to create commands and HCL/YAML files to copy-paste",
+	Long: `
+Helpers to create commands and HCL/YAML files to copy-paste. These commands are meant to make our life easier by creating copy-paste ready commands.
```
freshteapot, line 14: With the large PR #59 these commands are actually used by the system (https://github.com/dolittle/platform-api/tree/main/pkg/platform/job/k8s). I raise this point, as these commands are not just for the "platform-engineer".

Should we re-work the description to make that clear?
With the above knowledge, does "helpers" make sense?

freshteapot (changes requested):
Nice start @joelhoisko.
The helper functions that create jobs, need fixing in other parts of the code, as currently it is breaking.

- https://github.com/dolittle/platform-api/blob/main/pkg/platform/job/k8s/application.go#L15
- https://github.com/dolittle/platform-api/blob/main/pkg/platform/job/k8s/customer.go#L17

I am happy to adapt to using explore and helpers, I will have to consult you going forward to try and learn when to put them under "helpers" and "explore" vs other parts.

I see in the asana task there are sub-tasks around updating the k8s file etc, this will need to happen for the server.
Also, it is not documented very well (landed via #59) but you will also need to update the operations image environment variable as well as the deployment image.

Once we are closer to deploying, I am happy to watch you try and acknowledge all the missing documentation and improve it.

cmd/tools/explore/studio-customers.go
```diff
@@ -12,8 +12,8 @@ import (
 	"github.com/spf13/viper"
 )
 
-var getCustomersCMD = &cobra.Command{
-	Use:   "get-customers",
+var studioCustomersCMD = &cobra.Command{
```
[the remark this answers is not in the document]
joelhoisko: > My initial thinking around "explore" would be throw away commands to explore, maybe whilst developing or before we find value in them.

Okay I see then my confusion. When looking at the commands I only saw that they were just getting data in a structured way for us to explore or potentially to use in other commands and that they were read-only commands without side effects on the platform. So then it made sense to group these different "getters" of data into a single point and if you'd want to make another command that gets/exports data from the platform it would make sense for it to live there. So I didn't get that their point was to be "exploratory", just that they were a means for us to explore the data and to pipe it to other tools.

 > If I was new to the code base, I would think to find studio customers unders studio, not explore.

I agree, I also wasn't 100% sure on whether the `customers.json` getter command should live. I wanted to have it with the other "gettery" and read-only commands in `explore`, but it has more context while living in the `studio` command. Maybe it would then make sense to structure the `studio` subcommands like this:
- `studio`
  - `update` as all of these commands are about updating the existing `*.json` files.
    - `terraform`
    - `studio`
    - `application`
  - `create`
    - `service-account`
  - `get`
    - `customers`

This way we'd have different verbs on to act on the different subjects of Studio. The `studio admin` command could also be refactored into this structure over time. What do you think?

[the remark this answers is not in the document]
freshteapot: Let us do some brainstorming :).
See if it helps with the nice subcommands above, or if it confuses the matter.

Thinking about the jobs as they speak "application" and / or "customer" and less "technology".

cmd/tools/helpers/kubernetes/application-job.go
joelhoisko: I agree, I also don't like `helpers` or `utils` etc package names and I try to avoid them. Which is why I was also maybe thinking of calling this `templates` instead. I'm not sure that is there a nice word for what these commands do. Maybe `cli-templates`? The reason why I moved them into their own sub package was that I saw an opportunity to consolidate similar commands into one place. And I didn't like these helper commands names (they sounded like they were actually creating a customer or application), but thinking about it now maybe we should aim for more of the contextual "subject of the platform" first approach and then group things inside that subject.

So kinda like what I talked about the `studio` command, we could do something like:
- `job`
  -  `status` (so move the thing back from `explore`). We could also have the whole `job get status` thing going on, but I'm not sure that are there other things to "get" from a job than it's status.
  - `create` and in the future we could also use this to actually create jobs instead of just templates for the jobs
    - `customer-template`
    - `application-template`
  - OR
  - `template` as in the verb or we could be crazy and use ["templatize"](https://english.stackexchange.com/a/520300)
  	- `customer`
  	- `application`

The one downside I see with going with the `platform tools <subject> <verbs>` approach is that we're going to have a lot more subjects than verbs. For example in this approach the `explore` command isn't that different from "helpers" or "utils". Instead we'd want to have `platform tools ingress <get/list/something>`. And sometimes it's not that easy to figure out what the subject would be. What is the subject `import-dolittle-configmaps` is changing? I guess `kubernetes` or `platform` or maybe `configmaps`. And `pull-dolittle-config`'s subject is ultimately the Operations repo so maybe it could live in `platform tools <operations/repo> update dolittle-config`. And maybe the hardest one would be the `delete-application` and `delete-customer` commands as they ultimately are templates for the platform. So maybe `platform template delete application`? Or `platform tools platform template delete application`? But maybe at that point we could also get rid of the `tools` subcommand too...

I feel like that after having writing all of this I got some new ideas along the lines of `platform <subject> <verb>`ifying the CLI a bit more. But to get back to the task at hand, I propose this command structure (I'm also okay with `templatize`):
- `platform tools job status`
- `platform tools job template customer/application`
- `platform tools terraform template application`
- `platform tools terraform template customer`
- `platform template delete application`
- `platform template delete customer`
- Get rid of `helpers`

And also I didn't know that the job template commands were used in the jobs, I can fix that too when we land on the commands names.

joelhoisko: Implemented my proposal with the `template` verb.

cmd/tools/explore/studio-customers.go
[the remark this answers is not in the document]
joelhoisko: Implemented the changes to `studio`.

cmd/tools/helpers/root.go
joelhoisko: I fixed the created jobs to use the new command structure.

cmd/tools/job/template/customer.go
```diff
@@ -14,11 +14,11 @@ import (
...
 
-var createCustomerCMD = &cobra.Command{
-	Use:   "create-customer",
-	Short: "Create job to make customer",
+var customerCMD = &cobra.Command{
+	Use:   "customer",
+	Short: "Create a k8s Job to make customer",
 	Long: `
-	Outputs a new Dolittle platform customer in hcl to stdout.
+	Outputs a k8s job to create a customers application
```
freshteapot, line 21:
```suggestion
	Outputs a k8s job to create a customer
```

cmd/tools/studio/get/customers.go
```diff
@@ -12,17 +12,17 @@ import (
...
 	Long: `
-	Attempts to get customer.json
+	Gets the customer.json
 
 	GIT_REPO_BRANCH=dev \
 	GIT_REPO_DRY_RUN=true \
 	GIT_REPO_DIRECTORY="/tmp/dolittle-local-dev" \
 	GIT_REPO_DIRECTORY_ONLY=true \
-	go run main.go tools studio get-customers
+	go run main.go tools explore studio-customers
```
freshteapot, line 25:
```suggestion
	go run main.go tools studio get customers
```

cmd/tools/studio/update/application.go
```diff
@@ -20,17 +20,17 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-var buildApplicationInfoCMD = &cobra.Command{
-	Use:   "build-application-info",
-	Short: "Write application info into the git repo",
+var applicationCMD = &cobra.Command{
+	Use:   "application",
+	Short: "Update the application.json",
```
freshteapot, line 25: Not sure if it matters, but it also creates, not just updates.

cmd/tools/terraform/template/customer.go
```diff
@@ -19,13 +19,13 @@ type tfCustomer struct {
 	PlatformEnvironment string
 }
 
-var createCustomerHclCMD = &cobra.Command{
+var customerCMD = &cobra.Command{
 	Use:   "create-customer-hcl",
```
freshteapot, line 23: Does this need changing to match application?

cmd/tools/terraform/template/customer.go
```diff
@@ -19,13 +19,13 @@ type tfCustomer struct {
...
 	Short: "Write terraform output for a customer",
 	Long: `
 	Outputs a new Dolittle platform customer in hcl to stdout.
 
-	go run main.go tools terraform create-customer-hcl --name="Test Marka" --platform-environment="dev"
+	go run main.go tools terraform template customer --name="Test Marka" --platform-environment="dev"
```
freshteapot, line 28: The "use" doesnt match.

pkg/platform/job/k8s/application.go
```diff
@@ -114,7 +114,7 @@ func CreateApplicationResource(config CreateResourceConfig, customerID string, a
 						terraformApply(terraformBaseContainer, terrformFileName),
 						// Terraform create azure.json
 						terraformOutputJSON(terraformBaseContainer),
-						toolsStudioBuildTerraformInfo(platformImage, platformEnvironment, customerID),
+						toolsStudioUpdateTerraform(platformImage, platformEnvironment, customerID),
 						gitUpdateStudioTerraformInfo(platformImage, platformEnvironment, customerID, branch),
```
freshteapot, line 118: Do you want to change the name of this to be inline with the change above?

pkg/platform/job/k8s/customer.go
```diff
@@ -86,7 +86,7 @@ func CreateCustomerResource(config CreateResourceConfig, customer dolittleK8s.Sh
 							"sh",
 							"-c",
 							fmt.Sprintf(`
-/app/bin/app tools terraform create-customer-hcl \
+/app/bin/app tools terraform template customer \
```
freshteapot, line 89: This should have failed, based on the comment further up, or I am blind :).

freshteapot (changes requested):
Its looking good @joelhoisko.
I have spotted a few minor issues, bu the one main issue to clarify is 

```sh
/app/bin/app tools terraform template customer
```

Is not currently defined

```sh
go run main.go tools terraform template
Create pre-filled Terraform HCL

Usage:
  platform tools terraform template [command]

Available Commands:
  application         Write terraform output for a customer application
  create-customer-hcl Write terraform output for a customer

Flags:
  -h, --help   help for template

Global Flags:
      --platform-environment string   Platform environment (dev or prod), not linked to application environment (default "dev")

Use "platform tools terraform template [command] --help" for more information about a command.
```

I am also reminded, I haven't shared the secret for the service principle nor documented how to run this locally. (maybe you got it from the cluster :)).

cmd/tools/terraform/template/customer.go
joelhoisko: I fixed the command to be `tools terraform template customer`

cmd/tools/studio/update/application.go
joelhoisko: True, I guess it should be seen as an upsert, as should `platform tools studio update terraform` as well. And the `platform tools studio update studio` command is actually an "resetsert" or maybe just a reset then (I guess it's okay that a "reset" also creates)?

So we could have:
- `studio`
	- `upsert`
		- `terraform`
		- `application`
	- `reset`
		- `studio`

It makes sense, I just feel bad for dealing with this 3 days after tech debt day. I'll fix them.

joelhoisko: Actually I'm not 100% sure if `reset studio` is the correct term. Reading the code it's point seems to be to create default studio configs and/or reset existing ones to default. But you can also configure the command with flags to update/create a non-default studio config so it doesn't really work as "reset", even though that kinda was it's original idea.

So maybe it should live in `upsert`. I'll fix it.

joelhoisko: I fixed them all to be `upsert` and also fixed the jobs and their methods name to use the word "Upsert" instead of "Update".

cmd/tools/terraform/template/customer.go
joelhoisko: I assume you're talking about when I forgot to rename the `Use:` command to `customer`? Or what do you mean "match application"? The application command variable is called `applicationCMD` so it makes sense that the customer command variable is called `customerCMD`.

pkg/platform/job/k8s/application.go
joelhoisko: I'm not sure which of the methods you're talking about. I've already renamed the `toolsStudioUpdateTerraform()` to `toolsStudioUpsertTerraform()`. For the `gitUpdateStudioTerraformInfo()` method I don't think I care enough to go renaming everything right now.

pkg/platform/job/k8s/customer.go
joelhoisko: You weren't blind. Fixed.

freshteapot (approved):
🚢 🇮🇹 
Confirmed in dev :), made a customer and an application.

PR #95, opened 2022-03-02
Add a http handler for getting a customers studio.json

joelhoisko:
## Summary

Adds new endpoints:
- GET `/studio/customer/{customerID}` for fetching a single customer's `studio.json` file. The configuration in the response is camelCased so that it's nicer to consume in JavaScript/TypeScript.
- POST `/studio/customer/{customerID}` for updating a single customer's `studio.json` file. The request should be camelCased.

Example request payload for disabling all environments:
```json
{"buildOverwrite":true,"disabledEnvironments":["*"],"canCreateApplication":true}
```

Curl example:
```sh
echo -n '{"buildOverwrite":true,"disabledEnvironments":[],"canCreateApplication":true}' \ 
| curl -X POST localhost:8081/studio/customer/CUSTOMER_ID \
 -H "x-shared-secret: FAKE" \
 -H "user-id: DEVELOPMENT_USER_ID" \
 -H "tenant-id: DEVELOPMENT_CUSTOMER_ID" \
 -H "Content-Type: application/json" \
 --data-binary @-
```

These endpoints enable us to modify the `studio.json`, so that we can enable or disable the creation of microservices per customer more easily.

## Reference
- [Related Studio PR](https://github.com/dolittle/Studio/pull/159)
- [Asana](https://app.asana.com/0/1201885260673030/1201894099742054)


pkg/platform/studio/service.go
```diff
@@ -0,0 +1,99 @@
...
+
+func (s *service) Get(w http.ResponseWriter, r *http.Request) {
+	vars := mux.Vars(r)
+	customerID := vars["customerID"]
+	logContext := s.logContext.WithFields(logrus.Fields{
+		"customer_id": customerID,
+		"method":      "Get",
+	})
+
+	studioConfig, err := s.storageRepo.GetStudioConfig(customerID)
```
freshteapot, line 45: Over in the customer section, we use "hasAcess".
Do you think we should use has access here to reduce access?

pkg/platform/studio/service.go
```diff
@@ -0,0 +1,99 @@
...
+		return
+	}
+
+	studioConfig := platform.StudioConfig{
+		BuildOverwrite:       config.BuildOverwrite,
+		DisabledEnvironments: config.DisabledEnvironments,
+		CanCreateApplication: config.CanCreateApplication,
+	}
+
+	err := s.storageRepo.SaveStudioConfig(customerID, studioConfig)
```
freshteapot, line 88: Over in the customer section, we use "hasAcess".
Do you think we should use has access here to reduce access as it is a write operation?

pkg/platform/studio/service_test.go
```diff
@@ -0,0 +1,157 @@
...
+			}
+			jsonConfig = HTTPStudioConfig{
+				BuildOverwrite:       false,
+				DisabledEnvironments: []string{"*"},
+				CanCreateApplication: false,
+			}
+		})
+
+		It("returns a single studio configuration", func() {
+			request, _ := http.NewRequest("GET", customerUrl, nil)
```
freshteapot, line 60: I cant give very detailed reasons, but it seems "better" to use httptest here, what do you think?
Or why are you using http.NewRequest?

ie

```go
req = httptest.NewRequest("GET", url, nil)
```
there are examples in the tests which do this.

pkg/platform/studio/service_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+			mockRepo.On(
+				"GetStudioConfig",
+				customerID,
+			).Return(studioConfig, nil)
+
+			router.ServeHTTP(recorder, request)
```
freshteapot, line 67: I like what you have done here, what I did elsewhere is a little more clunky.

freshteapot (changes requested):
Nice work @joelhoisko, will follow up after my next 2 meetings.

- What to do about hasAccess
- can you run "go mod tidy", when I run it, it is different
- Love the tests, nice to see the coverage rise

I haven't tested it yet with the frontend, just read the code.

pkg/platform/studio/service_test.go
joelhoisko: I changed to use the `httptest.NewRequest()`, also found [this SO comment about it's intended use](https://stackoverflow.com/a/45682537/5806412).

pkg/platform/studio/service.go
joelhoisko: Good point, I'll update the tests and add a check for it

joelhoisko: I moved the `hasAccess()` method from the `customer` package to `k8s.Repo` package's `RepoRoleBinding` interfaces `HasUserAdminAccess(userID string) (bool, error)` method as it felt the closest place to put it without creating a whole new interface and struct for it in side `customer` package. But I also get if the `k8s.Repo` package is supposed to only deal in k8s resources instead of our own domain.

I fixed it to work with the customer service and with the studio and tested the studio parts locally (with the `PLATFORM_ENVIRONMENT="dev"` env variable on platform-api).

pkg/platform/studio/service.go
```diff
@@ -0,0 +1,127 @@
...
+	utils.RespondWithJSON(w, http.StatusOK, httpConfig)
+}
+
+func (s *service) Save(w http.ResponseWriter, r *http.Request) {
+	userID := r.Header.Get("User-ID")
+	vars := mux.Vars(r)
+	customerID := vars["customerID"]
+	logContext := s.logContext.WithFields(logrus.Fields{
+		"customer_id": customerID,
+		"method":      "Get",
```
freshteapot, line 84: Being the save func, I think we want save :)

```suggestion
		"method":      "Save",
```

freshteapot (changes requested):
@joelhoisko 1 minor change and then its good :).

I can't quite decide if we need "studio" as its own api endpoint, but happy to find out.
I have been flip flopping between the url you have created and something

> /customer/{customerID}/studoconfig

> /customer/{customerID}/config

And then having an attribute for studio.

I don't think it really matters as we try and figure it out :).

PR #98, opened 2022-03-08
Bug git pull from the side

freshteapot:
## Summary

Fix bug when running "api update-repo".


pkg/platform/storage/git/git_storage.go
```diff
@@ -249,3 +249,41 @@ func (s *GitStorage) IsAutomationEnabledWithStudioConfig(studioConfig platform.S
 func (s *GitStorage) GetRoot() string {
 	return filepath.Join(s.Directory, s.config.PlatformEnvironment)
 }
+
```
joelhoisko, line 252: I don't really understand why we are doing this, could you maybe add a comment explaining why we need to do this sort of file watching and then only pulling?

joelhoisko (changes requested):

pkg/platform/storage/git/git_storage.go
```diff
@@ -249,3 +249,41 @@ func (s *GitStorage) IsAutomationEnabledWithStudioConfig(studioConfig platform.S
...
+func (s *GitStorage) watchAndPull() {
```
freshteapot, line 253: @joelhoisko via "update-repo" command, it is possible to manually update the repo.
The problem is, it is broken.

1) Setup  the server
2) make a change in the repo
3) manually update the server via the command
4) do something by the server and observe it breaks working with git


Part of the reason I believe it is broken is the different way we "bootstrap" the worktree. Via NewGitStorage we have a few code paths.

We bootstrap the 1 version, where it interacts with the remote and downloads the repo and maintains it's state.
When we run the manual command, it doesn't interact with the remote and starts from an open repo, which is a slightly different state.

In the bowels of go-git, inside worktree.go PullContext the "fetch" function is explicit and doesn't include the RefSpecs, this might be part of the problem, not 100%. But seeing it is in part of the code we cant control.

I made some head way, but the moment I explored "Depth" or "SingleBranch" in both places, it broke due with "object not found".

Which in turn again bubbled up the thought, let me signal to the first "git storage" to pull, which has been shown to work, based on the jobs listener (https://github.com/dolittle/platform-api/blob/main/pkg/platform/job/applicationlistener.go#L81 there is a customer one too).

I opted to go for a way to signal to the "git storage" that was setup with the remote to pull.

Whilst writing this, I acknowledge

- I have not explored EnableDotGitCommonDir, this might create .git.
- Nor have I explored "recreating the "PlainCloneContext" to see if I can get the states to overlap.

I will add EnableDotGitCommonDir and see what happens.

freshteapot: Undoing my changes both via the command line and via the running server I get an error around

> err object not found

When I EnableDotGitCommonDir = true, everything is happy.
I do get a ".git" folder now.

I think using EnableDotGitCommonDir might be the better approach for now.

freshteapot: I was wrong, I forgot to cleanup :(.

freshteapot: When false or not set.

> {"branchReference":"refs/heads/bug-git-pull","context":"git-repo","directoryOnly":false,"error":"empty git-upload-pack given","gitBranch":"bug-git-pull","gitRemote":"git@github.com:dolittle-platform/Operations.git","level":"error","method":"Pull","msg":"Pull","time":"2022-03-10T13:10:48+01:00"}
err empty git-upload-pack given

It does help to use enableDotGitCommonDir.

It is not possible to make "Depth" or "Singlebranch" work. both bubble up different errors

```go
		// Neither of the below work
		//Depth:         1,
		// err object not found (doesnt work with either approach)
		//SingleBranch: true,
		// err empty git-upload-pack given
```

joelhoisko (approved):

PR #107, opened 2022-04-04
Add support for V8 Runtime with backwards compatibility set to V7 by default

joelhoisko:
## Summary

When creating a new microservice that uses the [V8.0.0](https://github.com/dolittle/Runtime/releases/tag/v8.0.0) Runtime, we default to setting the `DOLITTLE__RUNTIME__EVENTSTORE__BACKWARDSCOMPATIBILITY__VERSION` env variable to `V7`. This is set through the `appsettings.json`, where the ASP.NET configuration system will pick it up just like it would be set in the env variable.
We default to `V7`, because to `V6` should only be used when upgrading a V6 Runtime to V8 and we don't currently support upgrading of the Runtime in the platform.

## Reference
- https://app.asana.com/0/1202023614957161/1202062723906543
- https://github.com/dolittle/Runtime/releases/tag/v8.0.0
- https://github.com/dolittle/Studio/pull/170

### DIFF
--- a/pkg/platform/microservice/simple/k8s/resources.go
+++ b/pkg/platform/microservice/simple/k8s/resources.go
@@ -37,6 +37,8 @@ func NewResources(
 	switch runtimeImage {
 	case "dolittle/runtime:6.1.0":
 		dolittleConfig = dolittleK8s.NewMicroserviceConfigmapV6_1_0(microservice, customerTenants)
+	case "dolittle/runtime:8.0.0":
+		dolittleConfig = 
 	case "none":
 		fallthrough
 	default:
--- a/pkg/platform/microservice/simple/k8s/resources.go
+++ b/pkg/platform/microservice/simple/k8s/resources.go
@@ -38,7 +38,7 @@ func NewResources(
 	case "dolittle/runtime:6.1.0":
 		dolittleConfig = dolittleK8s.NewMicroserviceConfigmapV6_1_0(microservice, customerTenants)
 	case "dolittle/runtime:8.0.0":
-		dolittleConfig = 
+		dolittleConfig = dolittleK8s.NewMicroserviceConfigmapV8_0_0(microservice, customerTenants)
 	case "none":
 		fallthrough
 	default:
--- a/pkg/platform/microservice/simple/k8s/resources_test.go
+++ b/pkg/platform/microservice/simple/k8s/resources_test.go
@@ -1,6 +1,7 @@
 package k8s_test
 
 import (
+	"encoding/json"
 	"fmt"
 
 	dolittleK8s "github.com/dolittle/platform-api/pkg/dolittle/k8s"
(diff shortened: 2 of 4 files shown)


pkg/platform/microservice/simple/k8s/resources.go
```diff
@@ -37,6 +37,8 @@ func NewResources(
 	switch runtimeImage {
 	case "dolittle/runtime:6.1.0":
 		dolittleConfig = dolittleK8s.NewMicroserviceConfigmapV6_1_0(microservice, customerTenants)
+	case "dolittle/runtime:8.0.0":
```
freshteapot, line 40: With this additional case.

I am wondering if the default behaviour needs to evole.
As to the "none".

Or filter what versions of the runtime we currently do support at the API endpoint level.

pkg/platform/microservice/simple/k8s/resources_test.go
```diff
@@ -134,6 +135,21 @@ var _ = Describe("Resources", func() {
 		})
 	})
 
+	Describe("Creating resources", func() {
```
freshteapot, line 138: What does the value of this describe bring?
The top level is "Resources", just wondering if there is a few more words to give it context.
The "Context within" is great :).

freshteapot (approved):
Loooks good @joelhoisko.
Tested it in dev (thanks for hooking up).

I have approved it and your welcome to 🚢 🇮🇹 (ship it).

I have a question about the default behaviour or being explicit and filtering the runtimes the platform supports. But I am happy to come back to that later and not slow down progress. I will ping in teams and let you decide :).

pkg/platform/microservice/simple/k8s/resources_test.go
joelhoisko: I was looking at the [ginkgo docs](https://onsi.github.io/ginkgo/#adding-specs-to-a-suite) and they had this example for their  example `Books` package:

```go
var _ = Describe("Books", func() {
  var foxInSocks, lesMis *books.Book

  BeforeEach(func() {
    // .. skipping for brevity
  })

  Describe("Categorizing books", func() {
    Context("with more than 300 pages", func() {
      It("should be a novel", func() {
        Expect(lesMis.Category()).To(Equal(books.CategoryNovel))
      })
    })

    Context("with fewer than 300 pages", func() {
      It("should be a short story", func() {
        Expect(foxInSocks.Category()).To(Equal(books.CategoryShortStory))
      })
    })
  })
})
```

So I was trying to copy the type of structure that ginkgo was doing with. In this case I was describing the ability to create resources in the V8 context, inspired by ginkgos own wording here:
> We use **container nodes** like `Describe` and `Context` to organize the different aspects of code that we are testing hierarchically. In this case we are describing our book model's ability to categorize books across two different contexts - the behavior for large books "With more than 300 pages" and small books "With fewer than 300 pages".

Also later the [ginkgo docs](https://onsi.github.io/ginkgo/#organizing-specs-with-container-nodes) say:
> You usually `Describe` different capabilities of your code and explore the behavior of each capability across different `Contexts`.

So it felt to me that `Describe` was a more apt word to use than `Context` as I was describing an ability. I also get that this thing kinda boils down to semantics too...

joelhoisko:
Merging this in now so that we can deploy this.

pkg/platform/microservice/simple/k8s/resources_test.go
freshteapot: All valid points, I am still curious of the value of a Describe within a Describe.
As the Top Level makes it clear we are in "Resources", then the Context is "Create Resource".

But you are right, it is semantics and also taste.

PR #109, opened 2022-04-07
Create application, environments can have custom ids

freshteapot:
## Summary

- When creating an application, it is possible to define customTenants per environment
- Creating applications uses ./modules/dolittle-application-with-resources.


freshteapot:
# Reference
- https://app.asana.com/0/0/1202080188665938/f

PR #111, opened 2022-04-20
Add fileshare creation when creating a new applications environments

joelhoisko:
## Summary

Summary of the PR here. The GitHub release description is created from this comment so keep it nice and descriptive.

Remember to remove sections that you don't need or use.

### Added

- Describe the added features

### Changed

### Fixed

### Removed

- Describe what was removed and why

### Security

### Deprecated

### DIFF
--- a/pkg/azure/azure.go
+++ b/pkg/azure/azure.go
@@ -92,3 +92,34 @@ func LatestX(accountName string, accountKey string, shareName string) (ListRespo
 		Files:       latest,
 	}, nil
 }
+
+// EnsureFileShareExists tries to create a fileshare with a default quota in the given storage account.
+// If the fileshare already exists it returns nil.
+func EnsureFileShareExists(accountName, accountKey, shareName string) error {
+	// Use your Storage account's name and key to create a credential object; this is used to access your account.
+	credential, err := azfile.NewSharedKeyCredential(accountName, accountKey)
+
+	if err != nil {
+		return err
+	}
+
+	pipeline := azfile.NewPipeline(credential, azfile.PipelineOptions{})
+	u, err := url.Parse(fmt.Sprintf("https://%s.file.core.windows.net/%s", accountName, shareName))
+	if err != nil {
+		return err
+	}
+
+	shareURL := azfile.NewShareURL(*u, pipeline)
+	ctx := context.Background()
+
+	// NOTE: Metadata key names are always converted to lowercase before being sent to the Storage Service.
+	// Therefore, you should always use lowercase letters; especially when querying a map for a metadata key.
+	if _, err := shareURL.Create(ctx, azfile.Metadata{}, 0); err != nil {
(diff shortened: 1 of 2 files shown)


joelhoisko:
Tested manually with some kubectl cowboying and it worked.

jakhog (approved):
LGTM

pkg/platform/job/k8s/application.go
```diff
@@ -172,7 +172,7 @@ func buildApplicationInCluster(platformImage string, platformEnvironment string,
 	envVars = append(envVars, envVarGitNotInUse()...)
 
 	return corev1.Container{
-		Name:            "build-application-in-cluster",
+		Name:            "create-application",
```
freshteapot, line 175: I don't know the why the name got changed, but it had a knock on effect.

I ask, as this small change breaks applicationlistener.go.

https://github.com/dolittle/platform-api/blob/main/pkg/platform/job/applicationlistener.go#L123

Which in turn never updates when the application has finished building.
It is not a big thing, but it is a thing.

pkg/azure/azure.go
```diff
@@ -92,3 +93,39 @@ func LatestX(accountName string, accountKey string, shareName string) (ListRespo
...
+		return err
+	}
+
+	shareURL := azfile.NewShareURL(*u, pipeline)
+	ctx := context.Background()
+
+	// NOTE: Metadata key names are always converted to lowercase before being sent to the Storage Service.
+	// Therefore, you should always use lowercase letters; especially when querying a map for a metadata key.
+	if _, err := shareURL.Create(ctx, azfile.Metadata{}, 0); err != nil {
+		if storageErr, ok := err.(azfile.StorageError); ok && storageErr.ServiceCode() == "ShareAlreadyExists" {
```
freshteapot, line 119: Nice

freshteapot (commented):
Nice work on the PR.

This changes slightly the abstraction of using the "application job" to create the separate parts.

Ie we now have mixed the azure and kuberentes step. 

Not the end of the world, but a subtle change.

pkg/platform/job/k8s/application.go
joelhoisko: I changed the name so that it matched the name of the CLI command ran, plus like you pointed out it's not only about the k8s cluster anymore.

For the applicationlistener this line:
```go
if !strings.HasPrefix(pod.Name, "create-application-") {
	return false
}
```

Doesn't it check the name of the pod, not the container? I can see that we call these job pods `create-application-*` and isn't this line just to filter them out? Before making my change in the code I searched for the `build-application-in-cluster` string from Operations/platform-api/Studio and didn't find any other references to it.

Or is there something more hidden that this container name change breaks in the application listener? When I was creating the application in Studio I was able to refresh the page and see that it got reported as done on the page after the job finished (I assume that's the work of the applicationlistener).

PR #112, opened 2022-04-22
Enable head container commands and args to be defined on microservice creation

joelhoisko:
## Summary

Adds a new `HttpInputSimpleCommand` struct to the `Extra` field on the request for microservice creation. If the fields aren't specified the `command` and `args` fields will be emitted from the created deployment as the properties have the `omitempty` tag.

## Reference
- https://github.com/dolittle-platform/Operations/pull/209
- https://app.asana.com/0/1202121266838773/1202159723413336

### DIFF
--- a/pkg/platform/microservice/simple/k8s/resources.go
+++ b/pkg/platform/microservice/simple/k8s/resources.go
@@ -87,6 +87,7 @@ func NewDeployment(microservice dolittleK8s.Microservice, extra platform.HttpInp
 
 	deployment := dolittleK8s.NewDeployment(microservice, headImage, runtimeImage)
 
+	// the head container should always be the first container
 	deployment.Spec.Template.Spec.Containers[0].Command = headCommand.Commands
 	deployment.Spec.Template.Spec.Containers[0].Args = headCommand.Args
 	deployment.Spec.Template.Spec.Containers[0].Ports[0].ContainerPort = extra.HeadPort
--- a/pkg/platform/microservice/simple/k8s/resources.go
+++ b/pkg/platform/microservice/simple/k8s/resources.go
@@ -83,9 +83,12 @@ func NewResources(
 func NewDeployment(microservice dolittleK8s.Microservice, extra platform.HttpInputSimpleExtra) *appsv1.Deployment {
 	headImage := extra.Headimage
 	runtimeImage := extra.Runtimeimage
+	headCommand := extra.Headcommand
 
 	deployment := dolittleK8s.NewDeployment(microservice, headImage, runtimeImage)
 
+	deployment.Spec.Template.Spec.Containers[0].Command = headCommand.Commands
+	deployment.Spec.Template.Spec.Containers[0].Args = headCommand.Args
 	deployment.Spec.Template.Spec.Containers[0].Ports[0].ContainerPort = extra.HeadPort
 	return deployment
 }
--- a/pkg/platform/microservice/simple/k8s/resources.go
+++ b/pkg/platform/microservice/simple/k8s/resources.go
@@ -88,11 +88,11 @@ func NewDeployment(microservice dolittleK8s.Microservice, extra platform.HttpInp
 	deployment := dolittleK8s.NewDeployment(microservice, headImage, runtimeImage)
 
 	// the head container should always be the first container so we can trust in that for now
-	deployment.Spec.Template.Spec.Containers[0].Ports[0].ContainerPort = extra.HeadPort
-	// the Command property is equal to Dockers ENTRYPOINT, check https://stackoverflow.com/a/66078726/5806412
-	deployment.Spec.Template.Spec.Containers[0].Command = headCommand.Commands
-	// the Args property is equal to Dockers CMD
-	deployment.Spec.Template.Spec.Containers[0].Args = headCommand.Args
+	headContainer := deployment.Spec.Template.Spec.Containers[0]
+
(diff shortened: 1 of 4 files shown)


joelhoisko:
You can test this by creating a microservice like this in your local cluster:

```sh
curl -XPOST localhost:8081/microservice \
-H 'x-shared-secret: FAKE' \
-H 'Tenant-ID: 453e04a7-4f9d-42f2-b36c-d51fa2c83fa3' \
-H 'User-ID: local-dev' \
-H 'Content-Type: application/json' \
-d '
{
  "dolittle": {
    "applicationId": "11b6cf47-5d9f-438f-8116-0d9828654657",
    "customerId": "453e04a7-4f9d-42f2-b36c-d51fa2c83fa3",
    "microserviceId": "c96ab893-7df5-4065-a105-930c2c264ac2"
  },
  "name": "Test",
  "kind": "simple",
  "environment": "Dev",
  "extra": {
    "headImage": "nginxdemos/hello:latest",
    "runtimeImage": "dolittle/runtime:8.0.0",
    "ingress": {
      "path": "/",
      "pathType": "Prefix"
    },
    "isPublic": true,
    "headCommand": {
      "commands": ["echo"],
      "args": ["this is a test echo message"]
    }
  }
}'
```

Then check the logs:
```sh
kubectl -n application-11b6cf47-5d9f-438f-8116-0d9828654657 logs dev-test-*
this is a test echo message
```

pkg/platform/microservice/simple/k8s/resources.go
```diff
@@ -83,10 +83,16 @@ func NewResources(
...
+	headCommand := extra.Headcommand
 
 	deployment := dolittleK8s.NewDeployment(microservice, headImage, runtimeImage)
 
+	// the head container should always be the first container so we can trust in that for now
 	deployment.Spec.Template.Spec.Containers[0].Ports[0].ContainerPort = extra.HeadPort
+	// the Command property is equal to Dockers ENTRYPOINT, check https://stackoverflow.com/a/66078726/5806412
+	deployment.Spec.Template.Spec.Containers[0].Command = headCommand.Commands
+	// the Args property is equal to Dockers CMD
+	deployment.Spec.Template.Spec.Containers[0].Args = headCommand.Args
```
jakhog, line 95: What happens if this is not specified in the JSON from Studio? Do they turn into empty arrays that override the things in the Dockerfile, or does it keep the defaults?

We need the defaults from the Dockerfile to be used if nothing is supplied.

jakhog (commented):
I have a question about how this deals with defaults.

jakhog:
Shouldn't this be a minor?

pkg/platform/microservice/simple/k8s/resources.go
joelhoisko: I also wondered about this and when I tested it without args it seemed to work fine (I didn't do a full on test with making my own Docker image with custom `ENTRYPOINT` and `CMD` stuff).

Looking at the k8s Containers properties we can see that the `Command` and `Args` parts are both serialized with `omitempty` tag:
```go
// Entrypoint array. Not executed within a shell.
// The docker image's ENTRYPOINT is used if this is not provided.
// Variable references $(VAR_NAME) are expanded using the container's environment. If a variable
// cannot be resolved, the reference in the input string will be unchanged. The $(VAR_NAME) syntax
// can be escaped with a double $$, ie: $$(VAR_NAME). Escaped references will never be expanded,
// regardless of whether the variable exists or not.
// Cannot be updated.
// More info: https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#running-a-command-in-a-shell
// +optional
Command []string `json:"command,omitempty" protobuf:"bytes,3,rep,name=command"`
// Arguments to the entrypoint.
// The docker image's CMD is used if this is not provided.
// Variable references $(VAR_NAME) are expanded using the container's environment. If a variable
// cannot be resolved, the reference in the input string will be unchanged. The $(VAR_NAME) syntax
// can be escaped with a double $$, ie: $$(VAR_NAME). Escaped references will never be expanded,
// regardless of whether the variable exists or not.
// Cannot be updated.
// More info: https://kubernetes.io/docs/tasks/inject-data-application/define-command-argument-container/#running-a-command-in-a-shell
// +optional
Args []string `json:"args,omitempty" protobuf:"bytes,4,rep,name=args"`
```

Then I did some reading about slices and `nil`s. I learnt that a `nil` slice is considered an empty slice (it's still `nil` so it is different from an instantiated empty slice). Then reading the docs on [`omitempty`](https://pkg.go.dev/encoding/json#Marshal) it says that empty slices are omitted from the output:
> The "omitempty" option specifies that the field should be omitted from the encoding if the field has an empty value, defined as false, 0, a nil pointer, a nil interface value, and any **empty array, slice**, map, or string.

So if the `Command` or `Args` properties are either `nil` or an empty slice they would not be serialized to json for k8s. This is also what I saw in my testing with the containers having no `command` or `args` properties on them. You can also see that none of our `microservice-deployment.yml` files have any `command` or `args` on the containers, the property is just missing.

joelhoisko:
> Shouldn't this be a minor?

It changes the "schema" of the "database", although it's only an addition and that's an internal thing. This also means that the response for fetching a microservices info has an extra property on it, I'm not sure that is modifying the API responses with new properties considered a breaking change. If the consumer is very strict about the responses data then it could be breaking, but that's self imposed by the consumer. So it makes sense that it's only a minor.

PR #116, opened 2022-05-02
Api to get container registry info for a customer

freshteapot:
## Summary

Query the platform-api for images and tags in the container registry offered for each customer

### Added

- Get images
- Get tags based on image name

### DIFF
--- a/cmd/api/server.go
+++ b/cmd/api/server.go
@@ -16,6 +16,7 @@ import (
 	"github.com/dolittle/platform-api/pkg/platform/backup"
 	"github.com/dolittle/platform-api/pkg/platform/businessmoment"
 	"github.com/dolittle/platform-api/pkg/platform/cicd"
+	"github.com/dolittle/platform-api/pkg/platform/containerregistry"
 	"github.com/dolittle/platform-api/pkg/platform/customer"
 	"github.com/dolittle/platform-api/pkg/platform/insights"
 	"github.com/dolittle/platform-api/pkg/platform/job"
@@ -197,6 +198,12 @@ var serverCMD = &cobra.Command{
 			gitRepo,
 			logrus.WithField("context", "studio-service"),
 			k8sRepoV2,
+		)
+
+		containerRegistryService := containerregistry.NewService(
+			gitRepo,
+			containerregistry.NewAzureRepo(logContext.WithField("context", "container-registry-azure")),
+			logContext.WithField("context", "container-registry-service"),
 		)
 
 		c := cors.New(cors.Options{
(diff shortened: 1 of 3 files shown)


freshteapot:
# How
## Query images
```sh
curl -i -H 'Content-Type: application/json' \
-H 'x-shared-secret: FAKE' \
-H 'Tenant-ID: 453e04a7-4f9d-42f2-b36c-d51fa2c83fa3' \
-H 'User-ID: local-dev' http://localhost:8081/containerregistry/images
```

## Query tags

```sh
curl -i -H 'Content-Type: application/json' \
-H 'x-shared-secret: FAKE' \
-H 'Tenant-ID: 453e04a7-4f9d-42f2-b36c-d51fa2c83fa3' \
-H 'User-ID: local-dev' http://localhost:8081/containerregistry/image/abc/tags
```

goeran:
@freshteapot It would be cool if we could start the platform-api with a cmd flag to use local container registry (k3d) for local dev. I  don't think it's a lot of extra work to achieve this. Then we'll break the dependency to azure when developing locally.

What do you think?

freshteapot:
@goeran before this lands, I will rework to use acr secret from the cluster to be inline with the other use of acr

freshteapot:
hmm the other code expects an applicationID, today we don't send that in.
I guess we could, but it is not required.

freshteapot:
@goeran This latest commit https://github.com/dolittle/platform-api/pull/116/commits/784f821ff24f8cbf40853f419bc9303dbb429a3b, brings

- /application/{applicationID} to the url, which I don't like, but it is required to find the acr as it is in the namespace that needs to know the applicationID.
- We still rely on the customer in storage to get the correct key, as we can't rely 100% on order or that the acr will only ever contain one mapping inside ".dockerconfigjson", it is a safe bet but it might break

Is this better than splitting than mixing "storage with kubernetes", I honestly do not know. It did add more code and didn't free us from having to know the storage info.

## Options

- Good enough, refactor studio to use new endpoint
- Revert
- Something else

goeran:
# How to get it up and running locally

- clone Operations repo into /tmp/dolittle_operations
- create a feature branch in the repo - to avoid changing something in main
- remember to use dryrun

### Copy data for Dolittle tenant
```
cp -r /tmp/dolittle_operations/Source/V3/platform-api/prod/453e04a7-4f9d-42f2-b36c-d51fa2c83fa3 /tmp/dolittle_operations/Source/V3/platform-api/dev
```

### Copy acr secret for Dolittle Customer tenant from k8s cluster
```
kubectl --context Cluster-Production-Three-admin -n application-11b6cf47-5d9f-438f-8116-0d9828654657 get secrets acr -oyaml | kubectl --context k3d-dolittle-dev apply -f -
```

PR #117, opened 2022-05-03
Only check if the m3 connector environment variable is set when selecting to use the M3 connector

pavsaund:
## Summary

Fix error when trying to create a base microservice without the M3Connector and the environment variable set.


freshteapot (approved):
Nice work!

The shame is real @pavsaund.

PR #122, opened 2022-05-04
Automated Kafka resource provisioning in Aiven

joelhoisko:
## Summary

Adds a new CLI command `platform tools m3connector create environment` that provisions the kafka resources needed for an m3connector setup for an environment.
```bash
Usage:
  platform tools m3connector create environment [flags]

Flags:
      --application-id string   The applications ID
      --customer-id string      The customers ID
      --environment string      The environment
  -h, --help                    help for environment

Global Flags:
      --aiven-api-token string        Aiven API token
      --aiven-project string          Aiven project
      --aiven-service string          Aiven service
      --platform-environment string   Platform environment (dev or prod), not linked to application environment (default "dev")
```

The kafka resources are:
- 4 topics:
  - `cust_<customer-id>.app_<application-id>.env_<environment>.m3connector.change-events` (3 replicas, unlimited retention)
  - `cust_<customer-id>.app_<application-id>.env_<environment>.m3connector.input` (3 replicas, unlimited retention)
  - `cust_<customer-id>.app_<application-id>.env_<environment>.m3connector.commands` (3 replicas, unlimited retention)
  - `cust_<customer-id>.app_<application-id>.env_<environment>.m3connector.command-receipts` (3 replicas, 7d retention)
- An Aiven (sub) service user. Aiven's max length for usernames is 64 so we format the username in the following way:
```go
environment := "dev"
serviceName := "m3connector"
shortCustomerID := strings.ReplaceAll(customerID, "-", "")[:16]
shortApplicationID := strings.ReplaceAll(applicationID, "-", "")[:16]
username := fmt.Sprintf("%s.%s.%s.%s", shortCustomerID, shortApplicationID, environment, serviceName)
```
- Adds ACL's between the topics and the user with `"readwrite"` permissions.

The command will currently fail if any of the resources already existed.


joelhoisko:
## Testing

1. You need to have an Aiven user. In aiven go to your profile -> authentication and create a new token. Also note down the aiven project and service.
2. Run the command filling the correct flags (preferable a non-existant customer and application):
```bash
go run main.go tools m3connector create environment --aiven-api-token <api-token> --customer-id <customer-id> --application-id <application-id> --environment <environment> --project <aiven-project> --service <aiven-service>
```
3. Check the aiven topics, users and ACL's created that they match.
4. Remove the created resources afterwards to keep our aiven setup clean.

joelhoisko:
I'm not so sure about my naming of the `m3connector` package. My idea was that it would work as the business/domain logic side of all things m3connector related, but maybe it doesn't belong to live in `pkg/platform/microservice/m3connector` but rather just `pkg/platform/m3connector` as while it does deal with a microservice, it's not actually doing anything with the k8s microservice stuff, just provisioning stuff for it. I guess this is a bigger question overall about our package structure and what the packages mean.

pkg/aiven/aiven.go
```diff
@@ -0,0 +1,100 @@
...
+// https://github.com/aiven/aiven-go-client/blob/31343720eb5c31fbe37fe2ac188daa435b99ee4c/client.go#L71
+func NewClient(apiToken, project, service string, logger logrus.FieldLogger) (*Client, error) {
+	userAgent := fmt.Sprintf("aiven-go-client/%s", aiven.Version())
+	logContext := logger.WithFields(logrus.Fields{
+		"context":    "aiven",
+		"user_agent": userAgent,
+	})
+	aivenClient, err := aiven.NewTokenClient(apiToken, userAgent)
+	if err != nil {
+		logContext.WithField("error", err).Error("failed to create the aiven client with token")
```
mhelleborg, line 27: Here it might make more sense to return this as an error? Now it will be logged both from the caller and here

pkg/aiven/aiven.go
```diff
@@ -0,0 +1,100 @@
...
+		logContext: logContext,
+	}, nil
+}
+
+func (c *Client) CreateUser(username string) error {
+	logContext := c.logContext.WithFields(logrus.Fields{
+		"method":   "CreateUser",
+		"username": username,
+	})
+	logContext.Debug("creating a user")
```
mhelleborg, line 43: Do we need both logging before and after? Might make sense to just log once, when we know if it was successful

pkg/platform/microservice/m3connector/m3connector.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	err = m.createTopicAndACL(inputTopic, int64(-1), username)
+	if err != nil {
+		return err
+	}
+	err = m.createTopicAndACL(commandTopic, int64(-1), username)
+	if err != nil {
+		return err
+	}
+	hours, _ := time.ParseDuration("168h")
```
mhelleborg, line 99: We could do `time.Hour * 168`, if we are hardcoding it anyways

pkg/platform/microservice/m3connector/m3connector.go
```diff
@@ -0,0 +1,128 @@
...
+func (m *M3Connector) createTopicAndACL(topic string, retentionMs int64, username string) error {
+	logContext := m.logContext.WithFields(logrus.Fields{
+		"method":       "createTopicAndACL",
+		"topic":        topic,
+		"retention_ms": retentionMs,
+		"username":     username,
+	})
+	err := m.kafka.CreateTopic(topic, retentionMs)
+	if err != nil {
+		logContext.WithField("error", err).Error("failed to create the topic")
```
mhelleborg, line 118: Log & return on every level means lots of duplicate logging :) format the errors with details and log on the top level instead, for more concise and readable logs ;)

pkg/platform/microservice/m3connector/m3connector_test.go
```diff
@@ -0,0 +1,301 @@
...
+		customer = "8f90ddec-a668-4995-8399-2af9a8731723"
+		application = "e0f6a2a4-c136-4a56-80fa-75ff4daa75d6"
+		environment = "test"
+		resourcePrefix = fmt.Sprintf("cust_%s.app_%s.env_%s.m3connector", customer, application, environment)
+		shortCustomer := "8f90ddeca6684995"
+		shortApplication := "e0f6a2a4c1364a56"
+		username = fmt.Sprintf("%s.%s.%s.m3connector", shortCustomer, shortApplication, environment)
+	})
+	Describe("Creating a new environmet", func() {
+		It("should fail without a customer", func() {
```
mhelleborg, line 41: Nice and concise test cases 🙂

pkg/aiven/aiven.go
joelhoisko: I've been thinking this a bit. It's nice to have logging on multiple levels so that you can pinpoint where problems arise and ofc to return the original error from Aiven.

So to achieve both of these things I feel like there are 2 different ways to go about it:

1. Log the original error coming from aiven in the `aiven.go` part and then return our own custom error for the user of this aiven abstraction. Then we couldn't bubble up the original error to the API level which could be nice sometimes, although I think that the API maybe shouldn't be returning the errors or it's dependencies to the consumers.
2. I don't really know how this works in go, but apparently you can wrap errors into a chain by implementing an `Unwrap()` method on our custom domain specific error. This way we could have both our own custom error, and still contain the inner error for other purposes. Or is this way overkill with us having to declare our own types for errors?

And then there's the 3rd option where we just return the Aiven error and don't do custom stuff.

Any thoughts?

pkg/aiven/aiven.go
joelhoisko: My idea with it was that if we log both at the beginning and the end we can more easily see if something fails inside the method (and we didn't catch that error and log it). Not like platform-api has too much logging going on for it now either (the current log level for both dev and prod is info so these wouldn't even show up currently).

pkg/aiven/aiven.go
mhelleborg: If we go with [%w formatting of the internal error](https://stackoverflow.com/questions/61283248/format-errors-in-go-s-v-or-w), we don't need to create a custom type. Then we can just add context and wrap the internal error with very little ceremony.

This would give us the advantages of option 2, and is very close to how other languages handles inner exceptions. Then we only need to log it once, and we have all the context there including the root cause.

joelhoisko: I've changed the `aiven` package to just return formatted errors and not log them so that they can be logged elsewhere in the code.

mhelleborg (approved):

PR #123, opened 2022-05-05
Introducing GetTag v2 in the Platform-api and added tests

goeran:
# Summary

In order to be able to show and sort by creation date on the tags list in Studio, this pull request adds the following changes:

* Adds a `GetImageTags` for getting tags, including tag [metadata](https://docs.microsoft.com/en-us/rest/api/containerregistry/tag/get-list?tabs=HTTP#tagattributesbase) (`digest`, `createTime`, `lastUpdateTime`, and `signed` from azure container registry by image name.

* Adds new REST enpoint: `/application/{applicationID}/containerregistry/image-tags/{imageName:.*}`
* Adds new mockery generated mocks and some rudimentary tests
* Upgrades go-version in github ci and workflows from `1.16.3` to `1.18.3`
* Upgrades base image in the Docker file from `golang:1.16.3-alpine3.13` to ` golang:1.18.3-alpine3.15`
* Upgrades `k8s.io/api`, `k8s.io/apimachinery`, and `k8s.io/client-go` from  `v0.20.4` to v0.24.2`

## Ticket
 https://app.asana.com/0/1202121266838773/1202551576364785


pkg/platform/k8s/foobar.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/dolittle/platform-api/pkg/platform"
+	"github.com/sirupsen/logrus"
+	authv1 "k8s.io/api/authorization/v1"
+	corev1 "k8s.io/api/core/v1"
+	networkingv1 "k8s.io/api/networking/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/client-go/rest"
+)
+
+type K8sPlatformRepo interface {
```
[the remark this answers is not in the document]
joelhoisko: My reasoning for moving it is because usually in go I would like to declare the interfaces by their consumers. So if there was a struct that would consume `K8sRepo` I would like that struct to be the one declaring the functions that it wants and then it's up to the interfaces implementor to implement those. So instead of "I'm an interface, these are my methods for you to consume" it's more of a "I'm an interface, I define the things that I need from somebody".

But in the case of `K8sRepo` there isn't a single consumer, so maybe the approach then is to keep the interface in it's own file. This interface doesn't really follow the idea of interfaces in go so...

pkg/platform/k8s/k8s_repo.go
```diff
@@ -553,6 +553,7 @@ func (r *K8sRepo) CanModifyApplication(customerID string, applicationID string,
 // Only works when we can use the namespace
 // TODO bringing online the ad group from microsoft will allow us to check group access
 func (r *K8sRepo) CanModifyApplicationWithResourceAttributes(customerID string, applicationID string, userID string, attribute authv1.ResourceAttributes) (bool, error) {
+	fmt.Println("Tryprtint")
```
joelhoisko, line 556:
```suggestion
```

Leftovers

pkg/platform/k8s/k8s_repo.go
```diff
@@ -562,6 +563,7 @@ func (r *K8sRepo) CanModifyApplicationWithResourceAttributes(customerID string,
 		},
 	}
 
+	fmt.Println("Foobarx")
```
joelhoisko, line 566:
```suggestion
```

Leftovers

pkg/platform/microservice/doc.go
```diff
@@ -0,0 +1 @@
+package microservice
```
joelhoisko, line 1: I assume this file was created by accident and it can be deleted maybe?

pkg/platform/microservice/rawdatalog/nats.go
```diff
@@ -194,7 +194,7 @@ func createNatsResources(namespace, environment string, labels, annotations labe
 								},
 							},
 							LivenessProbe: &corev1.Probe{
-								Handler: corev1.Handler{
+								ProbeHandler: corev1.ProbeHandler{
```
joelhoisko, line 197: I assume these changes are from updating the dependencies and I'm assuming that they still work. Not like we use this part of the code anyway so.

pkg/platform/storage/git/git_storage.go
```diff
@@ -31,6 +32,15 @@ type GitStorageConfig struct {
...
 }
+
+/*type GitStorageFoo interface {
+	Pull() error
+	CommitPathAndPush(path string, msg string) error
+	IsAutomationEnabledWithStudioConfig(studioConfig platform.StudioConfig, applicationID string, environment string) bool
+	GetRoot() string
+	GetRepository() *git.Repository
+}*/
+
```
joelhoisko, line 43:
```suggestion
```

Leftovers

pkg/platform/storage/git/git_storage.go
```diff
@@ -126,6 +136,10 @@ func NewGitStorage(logContext logrus.FieldLogger, gitConfig GitStorageConfig) *G
 	return s
 }
 
+func (s *GitStorage) GetDirectory() string {
+	return s.Directory
+}
```
joelhoisko, line 141: Is there a reason not to just use `s.Directory`? Or if we want to keep the `Directory` property private then we should also make it private (lowercase it), otherwise this encapsulation doesn't really do anything. Or did you have something in mind for tests down the line?

cmd/api/server_test.go
```diff
@@ -0,0 +1,200 @@
...
+	Describe("When fetch images from Container Registry", func() {
+
+		It("should return 403 Forbidden when x-shared-secret header is not set", func() {
+			expect(
+				request{path: paths["images"], secret: ""},
+				response{status: http.StatusForbidden},
+			)
+		})
+
+		It("should rertun 403 Forbidden when x-shared-secret header is invalid", func() {
```
joelhoisko, line 122:
```suggestion
		It("should return 403 Forbidden when x-shared-secret header is invalid", func() {
```

typo

cmd/api/server_test.go
```diff
@@ -0,0 +1,200 @@
...
+							"createdTime":    now.Format(time.RFC3339Nano),
+							"lastUpdateTime": now.Format(time.RFC3339Nano),
+							"signed":         false,
+						},
+					},
+				},
+			)
+		})
+
+		It("should rertun 403 Forbidden when x-shared-secret header is invalid", func() {
```
joelhoisko, line 193:
```suggestion
		It("should return 403 Forbidden when x-shared-secret header is invalid", func() {
```

typo

cmd/api/mocks/k8splatformrepo_mock.go
```diff
@@ -0,0 +1,130 @@
+package mocks
```
joelhoisko, line 1: You don't have to write the mocks yourself, if you have interfaces defined then you can just run `make build-mocks` to build them using `mockery`. Then you can use those mocks with the proper `testify` library that has all the stuff around stubbing, mocking calls, asserting they worked etc etc. You do have to install mockery on your computer manually sadly.

Or was there maybe something blocking you from using mockery? I can now see that this testing was probably the reason why you put the git code inside the `storage.Repo` interface so that you could get it to work.

Example on how to use the mocks (there's many other places too to check them):
https://github.com/dolittle/platform-api/blob/3ca8a74fc6df4810dc9547416ada48a71998fdbe/pkg/platform/microservice/m3connector/m3connector_test.go#L185-L225

cmd/api/git-test.go
```diff
@@ -29,7 +29,7 @@ var gitTestCMD = &cobra.Command{
 			gitRepoConfig,
 		)
 
-		dir := filepath.Join(gitRepo.Directory, "dev", "453e04a7-4f9d-42f2-b36c-d51fa2c83fa3")
+		dir := filepath.Join(gitRepo.GetDirectory(), "dev", "453e04a7-4f9d-42f2-b36c-d51fa2c83fa3")
```
joelhoisko, line 32: Same stuff here about the encapsulation of the `Directory` property as I talked in the comment around `gitRepo`

cmd/api/server_test.go
```diff
@@ -0,0 +1,200 @@
...
+		containerRegistryRepo.StubAndReturnTags(([]string{"latest", "v1"}))
+		containerRegistryRepo.StubAndReturnTags2([]containerregistry.ImageTag{{
+			Name:           "label1",
+			Digest:         "sha256:...",
+			CreatedTime:    now,
+			LastUpdateTime: now,
+			Signed:         false,
+		}})
+
+		logContext := logrus.StandardLogger()
```
joelhoisko, line 90:
```suggestion
		logContext, _ = logrusTest.NewNullLogger()
```

You can also use the null logger from this package (the returned type from `NewNullLogger()` might be wrong but you get the idea):
```go
	logrusTest "github.com/sirupsen/logrus/hooks/test"
```

cmd/api/server_test.go
```diff
@@ -0,0 +1,200 @@
...
+		containerRegistryRepo = &mocks.ContainerRegistryMock{}
+		containerRegistryRepo.StubAndReturnImages([]string{"helloworld"})
+		containerRegistryRepo.StubAndReturnTags(([]string{"latest", "v1"}))
+		containerRegistryRepo.StubAndReturnTags2([]containerregistry.ImageTag{{
+			Name:           "label1",
+			Digest:         "sha256:...",
+			CreatedTime:    now,
+			LastUpdateTime: now,
+			Signed:         false,
+		}})
```
joelhoisko, line 88: This setup could be moved inside the `Describe("When fetch images from Container Registry")` block, as it's specific to that testing. Same goes for the paths I think, just to keep things together.

cmd/api/server.go
```diff
@@ -50,437 +53,467 @@ var serverCMD = &cobra.Command{
...
+		stdChainBase.ThenFunc(containerRegistryService.GetImages),
+	).Methods(http.MethodGet, http.MethodOptions)
+
+	router.Handle(
+		"/application/{applicationID}/containerregistry/tags/{imageName:.*}",
+		stdChainBase.ThenFunc(containerRegistryService.GetTags),
+	).Methods(http.MethodGet, http.MethodOptions)
+
+	router.Handle(
+		"/application/{applicationID}/containerregistry/tags2/{imageName:.*}",
```
joelhoisko, line 482: Same thing here about the `tags2` naming. And shouldn't it be a `repoName` or something like that that is passed in here? Or maybe I have misunderstood the whole thing.

joelhoisko (changes requested):
Nice work, lots of changes to go through:

- How did you actually update this? Me and Jakob have both struggled with this in the past due to the `json-iter` package panicking or something and would be nice to know the fix. What were the commands that you ran to update the project?
- Have you tested this locally or against production? Have you tested that the docker images build and run correctly?

Nice to see you getting to code in the platform-api and with tests too 👍

pkg/platform/containerregistry/azurerepo.go
```diff
@@ -77,3 +78,40 @@ func (repo azureRepo) GetTags(credentials ContainerRegistryCredentials, image st
 
 	return *result.Tags, nil
 }
+
+func (repo azureRepo) GetTags2(credentials ContainerRegistryCredentials, image string) ([]ImageTag, error) {
```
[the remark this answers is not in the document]
eliasgs: I changed this to GetImageTags

pkg/platform/containerregistry/service.go
```diff
@@ -24,6 +25,19 @@ type HTTPResponseTags struct {
...
+	Digest         string    `json:"digest"`
+	CreatedTime    time.Time `json:"createdTime"`
+	LastUpdateTime time.Time `json:"lastUpdateTime"`
+	Signed         bool      `json:"signed"`
+}
+
+type HTTPResponseTags2 struct {
+	Name string     `json:"name"`
+	Tags []ImageTag `json:"tags"`
+}
```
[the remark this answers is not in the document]
eliasgs: Changed to HTTPResponseImageTags

pkg/platform/microservice/rawdatalog/nats.go
eliasgs: Yes, they have renamed some stuff. I'm assuming they still work as well - not sure how test this?

cmd/api/server.go
eliasgs: We get tags by image name, so this should be fine.

pkg/platform/microservice/rawdatalog/nats.go
joelhoisko: Tbh I wouldn't worry too much about it, not even sure if they have worked for a while.

joelhoisko (approved):
:shipit:

PR #126, opened 2022-05-11
Add handling for both `data` & `binaryData` properties in configmaps

joelhoisko:
## Summary

- Fixes a bug with only the `binaryData` property for config files configmaps being used for fetching config files filenames.
- Adds support for both utf8 & non-utf8 files being uploaded as the configs. If the code detects that the incoming config file isn't valid utf8 it's going to get saved to the `binaryData` property. Otherwise it's saved to `data`.

## Reference
- https://app.asana.com/0/1202121266838773/1202244401000776

### DIFF
--- a/pkg/platform/microservice/configFiles/configFiles_suite_test.go
+++ b/pkg/platform/microservice/configFiles/configFiles_suite_test.go
@@ -1,0 +1,13 @@
+package configFiles_test
+
+import (
+	"testing"
+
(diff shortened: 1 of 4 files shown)


joelhoisko:
Tested this on the dev environment and it works all right, the text files and "binary" files get set in their respective propeties in the configmap:

image

goeran (approved):
🚢 🇮🇹

PR #127, opened 2022-05-12
Fix `RemoveEntryFromConfigFiles()` method to delete from the data property too

joelhoisko:
## Summary

Fixes a bug with config files saved on the configmaps `data` property not being deleted in the `RemoveEntryFromConfigFiles()` method.

## Reference
- https://app.asana.com/0/1202121266838773/1202261003053932/f

### DIFF
--- a/pkg/platform/microservice/configFiles/repo.go
+++ b/pkg/platform/microservice/configFiles/repo.go
@@ -136,18 +136,13 @@ func (r k8sRepo) RemoveEntryFromConfigFiles(applicationID string, environment st
 
 	configmapName := platformK8s.GetMicroserviceConfigFilesConfigmapName(name)
 	configMap, err := r.k8sDolittleRepo.GetConfigMap(applicationID, configmapName)
-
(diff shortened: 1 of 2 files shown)


pkg/platform/microservice/configFiles/repo.go
```diff
@@ -136,18 +136,13 @@ func (r k8sRepo) RemoveEntryFromConfigFiles(applicationID string, environment st
...
 		return err
 	}
 
-	if len(configMap.BinaryData) == 0 {
-		logContext.WithField("error", err).Error("no entries in configmap")
-		return err
-	}
-
 	delete(configMap.BinaryData, key)
+	delete(configMap.Data, key)
```
pavsaund, line 145: I learned today that delete is a no-op and can safely be run without putting guards in place. Nice!

pavsaund (approved):

pavsaund (approved):
My basic testing locally tells me this works!
🚢 🇮🇹 !

PR #141, opened 2022-06-16
Command to copy env variables for a microservice from en env to a new env

goeran:
## Summary
This was implemented to speed up Platform Engineers time spent on copying microservices from an environment to a new environment. 

With this command it's possible to copy env vars for a microservice in a specific environment to a new environment.

## Example
```
go run main.go tools microservice copy environment \
    --application cde2e951-d40a-3548-8b45-64c0ded97940 \
    --microservice-name frontend2 \
    --from-env test \
    --to-env prod
```


cmd/tools/studio/cfg/copyenv.go
```diff
@@ -0,0 +1,85 @@
...
+	"fmt"
+	"os"
+	"os/exec"
+
+	"github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+var heyCMD = &cobra.Command{
```
joelhoisko, line 15:
```suggestion
var copyenvCMD = &cobra.Command{
```

I assume heyCMD wasn't planned on being the final name :D

cmd/tools/studio/cfg/copyenv.go
```diff
@@ -0,0 +1,85 @@
...
+		destinationEnvironment := viper.GetString("tools.studio.cfg.to-env")
+		application := viper.GetString("tools.studio.cfg.application")
+
+		sourceConfigMapName := fmt.Sprintf("%s-%s-env-variables", sourceEnvironment, sourceMicroserviceName)
+		namespace := fmt.Sprintf("application-%s", application)
+		logContext.WithFields(logrus.Fields{
+			"k8sNamespace":     namespace,
+			"k8sConfigmap":     sourceConfigMapName,
+			"microserviceName": sourceMicroserviceName,
+		}).Info("Begins to copy env var configmap")
```
joelhoisko, line 42:
```suggestion
		logContext = logContext.WithFields(logrus.Fields{
			"k8sNamespace":     namespace,
			"k8sConfigmap":     sourceConfigMapName,
			"microserviceName": sourceMicroserviceName,
		})
		logContext.Info("Begins to copy env var configmap")
```

This way you save the properties that are setup into the `logContext` for later use.

cmd/tools/studio/cfg/copyenv.go
```diff
@@ -0,0 +1,85 @@
...
+
+		configMapJson, err := kubeCtlGetCfgMap.Output()
+		if err != nil {
+			logContext.WithFields(logrus.Fields{
+				"k8sNamespace": namespace,
+				"k8sConfigMap": sourceConfigMapName,
+				"error":        err,
+			}).Fatal("Failed to get configmap")
+			return
+		}
```
joelhoisko, line 59:
```suggestion
		ctx := context.TODO()
		k8sClient, _ := platformK8s.InitKubernetesClient()
		configMap, err := k8sClient.CoreV1().ConfigMaps(namespace).Get(ctx, sourceConfigMapName, metav1.GetOptions{})
		if err != nil {
			logContext.Fatal(err)
		}
```

Suggestion code not exact and missing imports, but should give you the idea.

Instead of just using raw `exec` powers, I suggest to use the k8s go SDK instead. If this command stays inside the `studio` subcommand then it already has the flags setup for kubeconfig, then you only need to call initialize the k8s client in the command and start using it.
https://github.com/dolittle/platform-api/blob/f643513e96152fdaf0675542258cbfc914114c88/cmd/tools/studio/root.go#L35-L41

cmd/tools/studio/cfg/copyenv.go
```diff
@@ -0,0 +1,85 @@
...
+
+		kubectlApply.Start()
+		kubeCtlGetCfgMap.Run()
+		kubectlApply.Wait()
+
+		logContext.WithFields(logrus.Fields{
+			"k8sNamespace":            namespace,
+			"k8sSourceCondfigMap":     sourceConfigMapName,
+			"k8sDestinationConfigMap": destinationConfigMap,
+		}).Info("Successfully copied configmap for env variables")
```
joelhoisko, line 83:
```suggestion
		logContext.Info("Successfully copied configmap for env variables")
```

If you setup the properties earlier then you don't need to keep repeating them.

cmd/tools/studio/cfg/root.go
```diff
@@ -0,0 +1,32 @@
...
+	heyCMD.MarkPersistentFlagRequired("microservice-name")
+	viper.BindPFlag("tools.studio.cfg.microservice-name", heyCMD.PersistentFlags().Lookup("microservice-name"))
+
+	heyCMD.PersistentFlags().String("from-env", "", "The environment to copy from")
+	heyCMD.MarkPersistentFlagRequired("from-env")
+	viper.BindPFlag("tools.studio.cfg.from-env", heyCMD.PersistentFlags().Lookup("from-env"))
+
+	heyCMD.PersistentFlags().String("to-env", "", "they environment to copy to")
+	heyCMD.MarkPersistentFlagRequired("to-env")
+	viper.BindPFlag("tools.studio.cfg.to-env", heyCMD.PersistentFlags().Lookup("to-env"))
```
joelhoisko, line 31: You don't need to use the `PersistentFlags()`, you can just use `Flags()`, that way these flags don't become global to any subcommands

joelhoisko (changes requested):
Nice, here are my thoughts:

To make the commands semantics a bit more aligned with the other commands and their design I think there are 2 main options:
- Create a new structure and folders along the lines of `tools microservice copy environment`. This is because this command doesn't have anything to do with Studio really, the other commands inside the `studio` parent command deal with more Studio specific stuff like upserting the different json files that platform-api serves to Studio. Granted it's not always that clear cut, but in this case this command sounds like it does something to the `studio.json` file or the actual Studio deployments. What this command ultimately deals with to me sounds more like microservices, eg "copy this microservices environment variables from this environment to this environment". Maybe it could also have something to do with environments, but that might be a bit of a reach. So I propose:
  - New folder and subcommand `microservice` inside `tools`. In the future we can start placing other microservice specific commands in here
  - New folder and subcommand/verb `copy` inside `microservice`. Similar setup as we have with the `tools studio` command where we have different verbs that act in Studios "domain".
  - The actual command `environment`/`env`/`env-variables`/`environment-variables`, indicating that this command acts on the microservices environment variables. So a full command could look something like `tools microservice copy env`
- If the above is a bit too much work for now (this is a "one off" like command after all), I'd suggest to just create a new verb `copy` inside the `studio` subcommand and then have the `environment`/`env`/`env-variables`/`environment-variables` be the actual commands name. So it would look like `tools studio copy env --microservice-name --from-env --to-env`. 


Overall another point would be to try to add this copy thing into some type of abstraction in platform-api to be used later by some other pieces of the code, but that could happen later (and tests etc)

cmd/tools/studio/cfg/copyenv.go
goeran: Lol :) no. Renamed

cmd/tools/studio/cfg/copyenv.go
goeran: You are correct. I'll fix

cmd/tools/studio/cfg/root.go
goeran: Thanks for the input. Will fix

cmd/tools/studio/cfg/copyenv.go
goeran: Done

cmd/tools/studio/cfg/copyenv.go
goeran: @joelhoisko  I think this is a good idea, but I don't think I want to spend more time on it. I think we can write this off as tech debt. When we incorporate copy of env variables into the platform-api we can rewrite it. Make sense?

cmd/tools/microservice/copy/root.go
```diff
@@ -0,0 +1,32 @@
...
+	environmentCMD.MarkPersistentFlagRequired("microservice-name")
+	viper.BindPFlag("tools.studio.cfg.microservice-name", environmentCMD.Flags().Lookup("microservice-name"))
+
+	environmentCMD.Flags().String("from-env", "", "The environment to copy from")
+	environmentCMD.MarkPersistentFlagRequired("from-env")
+	viper.BindPFlag("tools.studio.cfg.from-env", environmentCMD.Flags().Lookup("from-env"))
+
+	environmentCMD.Flags().String("to-env", "", "they environment to copy to")
+	environmentCMD.MarkPersistentFlagRequired("to-env")
+	viper.BindPFlag("tools.studio.cfg.to-env", environmentCMD.Flags().Lookup("to-env"))
```
joelhoisko, line 31:
```suggestion
	environmentCMD.Flags().String("application", "", "application id")
	environmentCMD.MarkFlagRequired("application")
	viper.BindPFlag("tools.microservice.copy.environment.application", environmentCMD.Flags().Lookup("application"))

	environmentCMD.Flags().String("microservice-name", "", "Name of microservice to copy environment variables from")
	environmentCMD.MarkFlagRequired("microservice-name")
	viper.BindPFlag("tools.microservice.copy.environment.microservice-name", environmentCMD.Flags().Lookup("microservice-name"))

	environmentCMD.Flags().String("from-env", "", "The environment to copy from")
	environmentCMD.MarkFlagRequired("from-env")
	viper.BindPFlag("tools.microservice.copy.environment.from-env", environmentCMD.Flags().Lookup("from-env"))

	environmentCMD.Flags().String("to-env", "", "they environment to copy to")
	environmentCMD.MarkFlagRequired("to-env")
	viper.BindPFlag("tools.microservice.copy.environment.to-env", environmentCMD.Flags().Lookup("to-env"))
```

The flags aren't persistent anymore so we can just use `MarkFlagRequired()`. These flags are also set on the `environmentCMD` only, so these flag/viper setups should be moved to the `environment.go` file's `init()` function so that they are self contained.

I also think that the viper strings should be changed to match the command, so `tools.microservice.copy.environment` etc. This is a minor thing because our viper configuration thing is all over the place anyways and in this case it's not even needed, you could do with just the `cobra` commands flags etc.

cmd/tools/microservice/copy/environment.go
```diff
@@ -0,0 +1,80 @@
...
+		kubectlApply.Stdin = bytes.NewReader(out)
+		kubectlApply.Stderr = os.Stderr
+
+		kubectlApply.Start()
+		kubeCtlGetCfgMap.Run()
+		kubectlApply.Wait()
+
+		logContextWithMeta.WithField("k8sDestinationConfigMap", destinationConfigMap).
+			Info("Successfully copied configmap for env variables")
+	},
```
joelhoisko, line 79:
```suggestion
		destinationConfigMap := fmt.Sprintf("%s-%s-env-variables", destinationEnvironment, sourceMicroserviceName)
		logContextWithMeta := logContext.WithFields(logrus.Fields{
			"k8sNamespace": namespace,
			"k8sConfigmap": sourceConfigMapName,
			"k8sDestinationConfigMap": destinationConfigMap,
			"microserviceName": sourceMicroserviceName,
		})

		logContextWithMeta.Info("Begins to copy env var configmap")

		kubeCtlGetCfgMap := exec.Command("kubectl", "-n", namespace, "get", "configmap", sourceConfigMapName, "-o", "json")
		_, err := kubeCtlGetCfgMap.StderrPipe()
		if err != nil {
			logContext.Fatal(err)
			return
		}

		configMapJson, err := kubeCtlGetCfgMap.Output()
		if err != nil {
			logContextWithMeta.WithField("error", err).
				Fatal("Failed to get configmap")
			return
		}

		var modifiedConfigMap map[string]interface{}
		json.Unmarshal(configMapJson, &modifiedConfigMap)
		metadata := modifiedConfigMap["metadata"].(map[string]interface{})
		metadata["name"] = destinationConfigMap
		// delete uid and version to ensure it patches
		delete(metadata, "resourceVersion")
		delete(metadata, "uid")
		out, _ := json.Marshal(modifiedConfigMap)

		kubectlApply := exec.Command("kubectl", "-o", "json", "apply", "-f", "-")
		kubectlApply.Stdin = bytes.NewReader(out)
		kubectlApply.Stderr = os.Stderr

		kubectlApply.Start()
		kubeCtlGetCfgMap.Run()
		kubectlApply.Wait()

		logContextWithMeta.Info("Successfully copied configmap for env variables")
	},
```

It makes the logging easier if you just setup all the properties from the start, that way you don't have to keep adding the fields temporarily later to a single log line.

cmd/tools/microservice/copy/environment.go
```diff
@@ -0,0 +1,80 @@
...
+	Use:   "environment",
+	Short: "Copy env variables configmap for a microservice in one environment to another environment",
+	Long: `
+	Copy env variables configmap for a microservice in one environment to another environment.
+
+	go run main.go tools studio cfg copyenv --application cde2e951-d40a-3548-8b45-64c0ded97940 \
+		--microservice-name frontend \
+		--from-env test \
+		--to-env prod
+	`,
```
joelhoisko, line 25:
```suggestion
	go run main.go tools microservice copy environment --application cde2e951-d40a-3548-8b45-64c0ded97940 \
		--microservice-name frontend \
		--from-env test \
		--to-env prod
	`,
```

Just fixing the old doc string

cmd/tools/microservice/copy/environment.go
joelhoisko, line 79:
```suggestion
		destinationConfigMap := fmt.Sprintf("%s-%s-env-variables", destinationEnvironment, sourceMicroserviceName)
		logContextWithMeta := logContext.WithFields(logrus.Fields{
			"k8sNamespace": namespace,
			"k8sConfigmap": sourceConfigMapName,
			"k8sDestinationConfigMap": destinationConfigMap,
			"microserviceName": sourceMicroserviceName,
		})

		logContextWithMeta.Info("Begins to copy env var configmap")

		kubeCtlGetCfgMap := exec.Command("kubectl", "-n", namespace, "get", "configmap", sourceConfigMapName, "-o", "json")
		_, err := kubeCtlGetCfgMap.StderrPipe()
		if err != nil {
			logContext.Fatal(err)
			return
		}

		configMapJson, err := kubeCtlGetCfgMap.Output()
		if err != nil {
			logContextWithMeta.WithField("error", err).
				Fatal("Failed to get configmap")
			return
		}

		var modifiedConfigMap map[string]interface{}
		json.Unmarshal(configMapJson, &modifiedConfigMap)
		metadata := modifiedConfigMap["metadata"].(map[string]interface{})
		metadata["name"] = destinationConfigMap
		// delete uid and version to ensure it patches
		delete(metadata, "resourceVersion")
		delete(metadata, "uid")
		out, _ := json.Marshal(modifiedConfigMap)

		kubectlApply := exec.Command("kubectl", "-o", "json", "apply", "-f", "-")
		kubectlApply.Stdin = bytes.NewReader(out)
		kubectlApply.Stderr = os.Stderr

		kubectlApply.Start()
		kubeCtlGetCfgMap.Run()
		kubectlApply.Wait()

		logContextWithMeta.Info("Successfully copied configmap for env variables")
	},
```

It makes the logging easier if you just setup all the properties from the start, that way you don't have to keep adding the fields temporarily later to a single log line.

joelhoisko, line 79:
```suggestion
		destinationConfigMap := fmt.Sprintf("%s-%s-env-variables", destinationEnvironment, sourceMicroserviceName)
		logContextWithMeta := logContext.WithFields(logrus.Fields{
			"k8sNamespace": namespace,
			"k8sConfigmap": sourceConfigMapName,
			"k8sDestinationConfigMap": destinationConfigMap,
			"microserviceName": sourceMicroserviceName,
		})

		logContextWithMeta.Info("Begins to copy env var configmap")

		kubeCtlGetCfgMap := exec.Command("kubectl", "-n", namespace, "get", "configmap", sourceConfigMapName, "-o", "json")
		_, err := kubeCtlGetCfgMap.StderrPipe()
		if err != nil {
			logContext.Fatal(err)
			return
		}

		configMapJson, err := kubeCtlGetCfgMap.Output()
		if err != nil {
			logContextWithMeta.WithField("error", err).
				Fatal("Failed to get configmap")
			return
		}

		var modifiedConfigMap map[string]interface{}
		json.Unmarshal(configMapJson, &modifiedConfigMap)
		metadata := modifiedConfigMap["metadata"].(map[string]interface{})
		metadata["name"] = destinationConfigMap
		// delete uid and version to ensure it patches
		delete(metadata, "resourceVersion")
		delete(metadata, "uid")
		out, _ := json.Marshal(modifiedConfigMap)

		kubectlApply := exec.Command("kubectl", "-o", "json", "apply", "-f", "-")
		kubectlApply.Stdin = bytes.NewReader(out)
		kubectlApply.Stderr = os.Stderr

		kubectlApply.Start()
		kubeCtlGetCfgMap.Run()
		kubectlApply.Wait()

		logContextWithMeta.Info("Successfully copied configmap for env variables")
	},
```

It makes the logging easier if you just setup all the properties from the start, that way you don't have to keep adding the fields temporarily later to a single log line.

joelhoisko: Not sure what the hell GitHub is reporting here as all these suggestions, but you get the point.

joelhoisko (changes requested):

cmd/tools/microservice/copy/root.go
goeran: @joelhoisko thanks. I will fix.

goeran: Agree with both of your suggestions.

cmd/tools/microservice/copy/environment.go
goeran: Agreed!

cmd/tools/microservice/copy/environment.go
goeran: I think you emphasised it well 😆 I'm on it

cmd/tools/microservice/copy/root.go
goeran: Now using MarkFlagRequired.

goeran: keys for viper also fixed.

cmd/tools/microservice/copy/environment.go
goeran: I agree, but it's not always the same fields.

cmd/tools/microservice/copy/environment.go
goeran: Fixed.

PR #145, opened 2022-10-13
Add network policy for platform router

woksin:
## Summary

Adds the network policy for the platform router so that it automatically added to new environments.

### Added

- Network policy for platform router


joelhoisko (approved):
:shipit:

PR #146, opened 2022-10-28
Add shell container to microservices

joelhoisko:
## Summary

Add a new container called `shell` with the `dolittle/studio-terminal` image to microservices when creating them.

Update `go.mod` to go 1.18


joelhoisko:
This PR should only be merged when we're ready to start rolling out the image to new microservices. This code also just puts the `latest` image onto the pods, which might not be what we want to do. On the other hand it could make it semi-auto-updating if it would always fetch the latest version of the image on each microservice restart.

PR #158, opened 2023-07-14
Document with diagrams

TomasEkeli:
## Summary

adds a go devcontainer -definition with 

- docker
- github-cli
- mockery
- k3d

adds diagrams showing the available routes

- docs/sitemap.md
- docs/routes_to_services.md

re-orders the routes in cmd/api/server.go to put all the routes under the same top-level path together


TomasEkeli:
i think we should create an open-api spec for this api.

